PR #6, opened 2024-05-23
Add deletion and finalizer logic for agentcontrolplane.

CrystalChun:
Adds the following behavior
- Adds finalizer to AgentControlPlane
- During delete of AgentControlPlane:
    - AgentControlPlane controller deletes ClusterDeployment 
    - Removes finalizer on AgentControlPlane

Deletion logic that exists already

- If an AgentControlPlane gets deleted, the following get auto-deleted too:
    - Machines
    - Metal3Machines (which deprovision the BMH)
    - AgentBootstrapConfig
- If a ClusterDeployment gets deleted, the AgentClusterInstall also gets removed


CrystalChun:
/retitle MGMT-17895: Add deletion and finalizer logic for agentcontrolplane

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -40,13 +41,15 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	controlplanev1beta1 "github.com/openshift-assisted/cluster-api-agent/controlplane/api/v1beta1"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 )
 
 const (
 	agentControlPlaneKind = "AgentControlPlane"
+	ACP_FINALIZER         = "agentcontrolplane." + controlplanev1beta1.Group + "/deprovision"
```
rccrdpccl, line 52: Please stick with CamelCase

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -93,12 +96,26 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		if rerr = r.Client.Status().Update(ctx, acp); rerr != nil {
 			log.Error(rerr, "couldn't update AgentControlPlane Status", "name", acp.Name, "namespace", acp.Namespace)
 		}
+		log.Info("Finished reconciling AgentControlPlane")
 	}()
 
+	if acp.DeletionTimestamp != nil {
+		log.Info("ACP is being deleted")
+		return r.handleDeletion(ctx, acp)
+	} else if !controllerutil.ContainsFinalizer(acp, ACP_FINALIZER) {
```
rccrdpccl, line 110: else is not necessary

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -199,6 +216,70 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	return ctrl.Result{}, rerr
 }
 
+// Ensures dependencies are deleted before allowing the AgentControlPlane to be deleted
+// Deletes the ClusterDeployment (which deletes the AgentClusterInstall)
+// Machines, InfraMachines, and AgentBootstrapConfigs get auto-deleted when the ACP has a deletion timestamp - this deprovisions the BMH automatically
+// TODO: should we handle watching until all machines & agentbootstrapconfigs are deleted too?
+func (r *AgentControlPlaneReconciler) handleDeletion(ctx context.Context, acp *controlplanev1beta1.AgentControlPlane) (ctrl.Result, error) {
```
rccrdpccl, line 223: This methods is quite long, can we extract `deleteClusterDeployment(...)` and `deleteInfraEnv(...)`?

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -199,6 +216,70 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			if err := r.Client.Get(ctx, types.NamespacedName{Namespace: acp.Spec.AgentConfigSpec.ClusterDeploymentRef.Namespace, Name: acp.Spec.AgentConfigSpec.ClusterDeploymentRef.Name}, cd); err != nil {
+				if !apierrors.IsNotFound(err) {
+					log.Error(err, "failed getting cluster deployment for ACP")
+					return ctrl.Result{}, err
+				}
+				// Continue if not found
+			} else if err := r.Client.Delete(ctx, cd); err != nil {
+				log.Error(err, "failed deleting cluster deployment for ACP")
+				return ctrl.Result{}, err
+			}
```
rccrdpccl, line 240: What's the downside of running delete directly (without running GET previously)? Can we differentiate "not found" on deletion from other errors?

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -93,12 +96,26 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 		return ctrl.Result{}, err
 	}
+	log.Info("Started reconciling AgentControlPlane")
```
rccrdpccl, line 99: I know we never cared so far, but I'd suggest to set logging right: is this Info or Debug?

controlplane/internal/controller/agentcontrolplane_controller.go
rccrdpccl: Btw, not sure we should handle deleting infraenv from this controller, as the infraenv is being created by AgentBootstrap
Wouldn't the approach of letting cluster deployment cascade deletion handle it work?

rccrdpccl:
Do you think we can start adding simple tests? In this case if we can add something that proves only deletion logic would be great.

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: How come? This only updates the CR if it doesn't already have the finalizer (we don't want it to update all the time), otherwise it skips the update

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: I tested it out, and was wrong - the clusterDeployment doesn't actually cascade delete the infraenv :(

CrystalChun: Also remember we talked about the AgentBootstrapConfig is not a one-to-one mapping to infraEnv so we can't just delete an infraEnv everytime we delete an AgentBootstrapConfig because of how it's currently designed. It would need to be by Control Plane or by Machine Deployment because those have a 1:1 mapping to infra envs

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: I was following what assisted-service [does](https://github.com/openshift/assisted-service/blob/66a3ef7405297ff5b3dd21feee1427f03f068dda/internal/controller/controllers/agent_controller.go#L116) but I have no preference, what do you think it should be?

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Good point! Modified to just delete

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Modified, thanks!

controlplane/internal/controller/agentcontrolplane_controller.go
rccrdpccl: if `DeletionTimestamp` is not nil, we will handle deletion and return.
This means we can remove the else, we can just have another if condition (`if !controllerutil.ContainsFinalizer(...)`).
I think this would make the code more readable.

controlplane/internal/controller/agentcontrolplane_controller.go
rccrdpccl: Ok, if we can't leverage cascade deleting, I'd propose the following:
* we set the finalizer on infraenv from agentbootstrapconfig
* we can index `bootstrapConfig.Status.InfraEnvRef` field
* on deletion, we can check if there is any other resource referencing such infraEnv, if not we can handle deletion

The advantages I can see of the above approach is the following:
* we keep the code related to creating and deleting infraenv in the same package/controller
* we don't need extra RBAC permissions in other controllers to handle infraenv

On the other hand, it'll be slightly more complicated than just sharing logic between the controllers as we are doing in this PR.

WDYT?

controlplane/internal/controller/agentcontrolplane_controller.go
rccrdpccl: Let's agree on some logging "rules":
* when to use what log level (i.e. when to use `INFO`, when to use `DEBUG`, etc)
* what information we should log

We can discuss this offline and come up with a strategy, feel free to ignore this comment for this PR.

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Oh I got you, that makes sense, thanks for explaining. Modified as suggested!

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Discussed with Riccardo offline and determined that 
1. we will review logging in each PR 
2. follow general guidelines:

    - log info when necessary to know and not repetitive
    - log debug levels if it's necessary for debugging
3. A refactor of all logging will be done later on
https://issues.redhat.com/browse/MGMT-17959

CrystalChun:
> Do you think we can start adding simple tests? In this case if we can add something that proves only deletion logic would be great.

Agreed! Added a couple of unit tests

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -85,20 +87,33 @@ type AgentControlPlaneReconciler struct {
...
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, err
 	}
+	log.WithValues("AgentControlPlane Name", acp.Name, "AgentControlPlane Namespace", acp.Namespace)
+	log.V(5).Info("Started reconciling AgentControlPlane")
 	defer func() {
 		if rerr = r.Client.Status().Update(ctx, acp); rerr != nil {
-			log.Error(rerr, "couldn't update AgentControlPlane Status", "name", acp.Name, "namespace", acp.Namespace)
+			log.V(3).Error(rerr, "couldn't update AgentControlPlane Status")
```
rccrdpccl, line 101: Shouldn't this be V(3) info? It seems strange to use verbosity for error

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -154,10 +165,6 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			// Cluster deployment no longer exists, unset reference and re-reconcile
 			acp.Status.ClusterDeploymentRef = nil
 			log.Info("Clusterdeployment doesn't exist, unset reference and reconcile again")
-			if err := r.Client.Update(ctx, acp); err != nil {
```
rccrdpccl, line 157: Why is this removed?

controlplane/internal/controller/agentcontrolplane_controller.go
rccrdpccl: Yeah, if you agree I would delegate InfraEnv handling to the bootstrap provider

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -199,6 +206,43 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			cd := &hivev1.ClusterDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      acp.Status.ClusterDeploymentRef.Name,
+					Namespace: acp.Status.ClusterDeploymentRef.Namespace,
+				},
+			}
+			if err := r.Client.Delete(ctx, cd); err != nil && !apierrors.IsNotFound(err) {
+				log.Error(err, "failed deleting cluster deployment for ACP")
+				return ctrl.Result{}, err
+			}
```
rccrdpccl, line 228: We can extract this into `deleteClusterDeployment(acp.StatusClusterDeploymentRef)`

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Ah yeah sure, I'll change it back to error - wasn't sure if it was that important to see all the time

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Oh yes sorry I forgot about this, will modify!

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -199,6 +206,43 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 }
 
+// Ensures dependencies are deleted before allowing the AgentControlPlane to be deleted
+// Deletes the ClusterDeployment (which deletes the AgentClusterInstall)
+// Machines, InfraMachines, and AgentBootstrapConfigs get auto-deleted when the ACP has a deletion timestamp - this deprovisions the BMH automatically
+// TODO: should we handle watching until all machines & agentbootstrapconfigs are deleted too?
+func (r *AgentControlPlaneReconciler) handleDeletion(ctx context.Context, acp *controlplanev1beta1.AgentControlPlane) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	if controllerutil.ContainsFinalizer(acp, acpFinalizer) {
+		log.Info("ACP contains finalizer, handling deletion")
```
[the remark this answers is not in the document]
CrystalChun: Good idea, modifying!

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Yep! Thanks Riccardo!

controlplane/internal/controller/agentcontrolplane_controller.go
rccrdpccl: Oh, you're right :sweat_smile:

controlplane/internal/controller/agentcontrolplane_controller.go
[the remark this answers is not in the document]
CrystalChun: Changed the logging

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Changed it!

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Changed it!

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -199,6 +206,45 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		acp.Status.ClusterDeploymentRef = nil
+
+		if controllerutil.RemoveFinalizer(acp, acpFinalizer) {
+			if err := r.Client.Update(ctx, acp); err != nil {
+				log.Error(err, "failed updating the ACP finalizer")
+				return ctrl.Result{}, err
+			}
+		}
+	}
+	log.Info("ACP doesn't contain finalizer, allow deletion")
```
rccrdpccl, line 231: This will be shown even when finalizer has just been removed. Also this would probably be better as debug, WDYT?

CrystalChun: Modified it!

PR #9, opened 2024-05-30
WIP: Add finalizer and deletion logic for agent bootstrapconfig

CrystalChun:
Kubeadm bootstrap just ignores any config being deleted https://github.com/kubernetes-sigs/cluster-api/blob/d1a230153b4bb27f434f6ccb3acbd630ce0979d7/bootstrap/kubeadm/internal/controllers/kubeadmconfig_controller.go#L237

### DIFF
--- a/bootstrap/internal/controller/agentbootstrapconfig_controller.go
+++ b/bootstrap/internal/controller/agentbootstrapconfig_controller.go
@@ -96,9 +96,6 @@ func (r *AgentBootstrapConfigReconciler) Reconcile(ctx context.Context, req ctrl
 	log := ctrl.LoggerFrom(ctx)
 
 	log.V(logutil.TraceLevel).Info("Reconciling AgentBootstrapConfig")
-	defer (func() {
-		log.V(logutil.TraceLevel).Info("Finished reconciling AgentBootstrapConfig")
-	})()
 
 	config := &bootstrapv1alpha1.AgentBootstrapConfig{}
 	if err := r.Client.Get(ctx, req.NamespacedName, config); err != nil {
@@ -167,10 +164,8 @@ func (r *AgentBootstrapConfigReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, fmt.Errorf("cluster name label not found in config")
 	}
 
-	// Get the Machine associated with this agentbootstrapconfig
-
-	// TODO: change the way we get this, for now it has the same name but that may not be the case - we should change to fetch by spec's reference to this agentbootstrapconfig
-	machine, err := util.GetMachineByName(ctx, r.Client, config.Namespace, config.Name)
+	// Get the Machine that owns this agentbootstrapconfig
+	machine, err := util.GetOwnerMachine(ctx, r.Client, config.ObjectMeta)
 	if err != nil {
 		log.Error(err, "couldn't get machine associated with agentbootstrapconfig", "name", config.Name)
 		return ctrl.Result{}, err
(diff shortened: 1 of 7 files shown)


bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -122,6 +121,12 @@ func (r *AgentBootstrapConfigReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 		return ctrl.Result{}, err
 	}
+
+	// Ignore any agent bootstrapconfig being deleted
+	if !config.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
```
CrystalChun, line 127: Theoretically, the metal3machine takes care of deprovisioning the BMH. Not sure what we should do about the agent? 
If it's a control plane node, we probably wouldn't allow detaching until the AgentControlPlane gets deleted...

bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -294,6 +312,28 @@ func (r *AgentBootstrapConfigReconciler) isReferencingACI(clusterDeployment *hiv
 	return clusterDeployment.Spec.ClusterInstallRef != nil
 }
 
+func (r *AgentBootstrapConfigReconciler) handleDeletion(ctx context.Context, config *bootstrapv1beta1.AgentBootstrapConfig, owner *bsutil.ConfigOwner) (ctrl.Result, error) {
```
CrystalChun, line 315: Theoretically, the metal3machine takes care of deprovisioning the BMH. Not sure what we should do about the agent?

rccrdpccl: Let's say a user wants to scale down, what will happen?
Replicas are adjusted either in control plane or machine deployment. Machine are going to be deleted, which in turn will probably delete metal3machine, which will delete the BMH.

What would happen to agentbootstrapconfig? Is machine controller going to issue a delete to agentboostrapconfig? If so, should we safely remove the node (agent) from the cluster?

I think it's reasonable to think of deletion as scaling down

rccrdpccl: As discussed offline we will focus on SNO: the machine deletion is going to cause BMH to get deprovisioned, so cleaning up the agent CR has no side effect.

bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -143,15 +149,24 @@ func (r *AgentBootstrapConfigReconciler) Reconcile(ctx context.Context, req ctrl
...
 	if !ok {
 		return ctrl.Result{}, fmt.Errorf("cluster name label not found in config")
 	}
 
-	// Get the Machine associated with this agentbootstrapconfig
-
-	// TODO: change the way we get this, for now it has the same name but that may not be the case - we should change to fetch by spec's reference to this agentbootstrapconfig
-	machine, err := util.GetMachineByName(ctx, r.Client, config.Namespace, config.Name)
+	// Get the Machine that owns this agentbootstrapconfig
+	machine, err := util.GetOwnerMachine(ctx, r.Client, config.ObjectMeta)
```
rccrdpccl, line 169: :+1:

bootstrap/main.go
```diff
@@ -38,6 +38,7 @@ import (
 
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	metal3 "github.com/metal3-io/cluster-api-provider-metal3/api/v1beta1"
+	v1beta12 "github.com/openshift-assisted/cluster-api-agent/controlplane/api/v1beta1"
```
rccrdpccl, line 41: should we consistently alias this imports to `controlplanev1beta1` and `bootstrapv1beta1` ?

rccrdpccl:
Looks good, could u please rebase with alpha1 version and add a test case? Feel free to merge it after that, if we have further commens we can open a new PR, but at least this goes in.

CrystalChun:
/retitle: MGMT-17895: Finalizers and deletion logic for agent bootstrap config controller

CrystalChun:
/retitle MGMT-17895: Finalizers and deletion logic for agent bootstrap config controller

PR #10, opened 2024-06-04
set CRDs version to v1alpha1

rccrdpccl:
Moving API to v1alpha1 to allow for change in the future


controlplane/api/v1alpha1/groupversion_info.go
CrystalChun, line 1: Same here changing the Version in the code to v1alpha1 
https://github.com/openshift-assisted/cluster-api-agent/pull/10/files#diff-11a782bc5a0182e58c917791a5e7b3fb7bfd5ec7c1c2ac1da68584590b8e413bR29

CrystalChun:
/retitle MGMT-18034: Set CRD versions to v1alpha1

PR #12, opened 2024-06-06
MGMT-17972: add tests to agentbootstrapconfig controller

rccrdpccl:
This PR is mainly adding tests, plus some small fixes.


rccrdpccl:
Merging so we can carry on with other work but please do comment - if anything comes up I will fix in another PR

PR #13, opened 2024-06-06
bugfix: add rbac and remove debug code

rccrdpccl:
The tests did not catch this because ACP would be found as RBAC do not come into play in unit-tests.


CrystalChun (approved):
/lgtm

PR #16, opened 2024-06-07
bugfix: map cluster API v1beta1 to agent provider v1alpha1 version

rccrdpccl:
This maps cluster API apiversion with our provider's, otherwise cluster API controller cannot select our resources.


CrystalChun:
/retitle MGMT-18034: bugfix: map cluster API v1beta1 to agent provider v1alpha1 version

PR #18, opened 2024-06-12
MGMT-17963: Enable pull-through cache setup

CrystalChun:
A more UX-friendly approach.

User provides the same configmap they would give to assisted service (src), e.g.:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: mirror-registry-config
data:
  ca-bundle.crt: |
    -----BEGIN CERTIFICATE-----
    certificate contents
    -----END CERTIFICATE-----

  registries.conf: |
    unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]

    [[registry]]
       prefix = ""
       location = "quay.io"
       mirror-by-digest-only = true

       [[registry.mirror]]
       location = "10.1.178.25:5000"

```

We convert all of it to this configmap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: image-config
data:
  image-config.yaml: |
    apiVersion: config.openshift.io/v1
    kind: Image
    metadata:
      name: cluster
    spec:
      registrySources:
        insecureRegistries:
        - 10.1.178.25:5000
      additionalTrustedCA:
        name: trusted-ca
  test-ca.yaml: |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: trusted-ca
      namespace: openshift-config
    data:
      registry.example.com: |
        -----BEGIN CERTIFICATE-----
        ...
        -----END CERTIFICATE-----
  idms.yaml: |
    apiVersion: config.openshift.io/v1
    kind: ImageDigestMirrorSet
    metadata:
      name: dockerio-mirror
    spec:
      imageDigestMirrors:
      - mirrors:
        - 10.1.178.25:5000
        source: quay.io
```


controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -251,6 +266,20 @@ func computeAgentClusterInstall(clusterDeployment *hivev1.ClusterDeployment, acp
 	if cluster.Spec.ClusterNetwork != nil && cluster.Spec.ClusterNetwork.Services != nil {
 		serviceNetwork = cluster.Spec.ClusterNetwork.Services.CIDRBlocks
 	}
+	var additionalManifests []hiveext.ManifestsConfigMapReference
+	if len(acp.Spec.AgentConfigSpec.ManifestsConfigMapRefs) > 0 {
+		additionalManifests = append(additionalManifests, acp.Spec.AgentConfigSpec.ManifestsConfigMapRefs...)
+	}
+
+	if acp.Spec.AgentConfigSpec.ImageRegistryRef != nil {
```
rccrdpccl, line 274: Can we extract image registry functionality to a different package, maybe `controlplane/internal/registry` or something similar? So that the body below can look something like:
```suggestion
	if acp.Spec.AgentConfigSpec.ImageRegistryRef != nil {
	
	     imageRegistryManifestRef, err := imageRegistry.CreateConfig(ctx, r.Client, acp.Spec.AgentConfigSpec.ImageRegistryRef)
	     
	     if err != nil {
	         // handle error
	     }
	     additionalManifests = append(additionalManifests, imageRegistryManifestRef)
	}
```

rccrdpccl:
Looks good! As commented within the code I would extract the functionality around creating the config to another package.

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Good idea, thank you! Will do!

controlplane/internal/controller/imageregistry/image_registry.go
```diff
@@ -0,0 +1,199 @@
...
+
+func createImageRegistryConfig(imageRegistry *corev1.ConfigMap, namespace string) (*corev1.ConfigMap, error) {
+	imageConfig := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      imageConfigMapName,
+			Namespace: namespace,
+		},
+		Data: make(map[string]string),
+	}
+	registryConf, ok := imageRegistry.Data[registryConfKey]
```
rccrdpccl, line 59: What if the key is not there?

controlplane/internal/controller/imageregistry/image_registry.go
```diff
@@ -0,0 +1,199 @@
...
+		mirrorTrees, ok := registryTree.Get("mirror").([]*toml.Tree)
+		if !ok {
+			continue
+		}
+
+		for _, mirrorTree := range mirrorTrees {
+			mirror, ok := mirrorTree.Get("location").(string)
+			if ok {
+				mirrors = append(mirrors, configv1.ImageMirror(mirror))
+			}
```
rccrdpccl, line 118: I think inlining it would be tidier
```if mirror, ok := mirrorTree.Get("location").(string); ok {
    mirrors = append(mirrors, configv1.ImageMirror(mirror))
}
```

controlplane/internal/controller/imageregistry/image_registry.go
```diff
@@ -0,0 +1,199 @@
...
+		},
+	}
+
+	idmsJSON, err := json.Marshal(idms)
+	if err != nil {
+		return "", err
+	}
+
+	return string(idmsJSON), nil
+}
```
rccrdpccl, line 151: I think this can be broken down into multiple smaller functions.

I see:
* getImageRegistries
* one (funcA) that wraps the logic returning `idmsMirrors`
* another one within funcA (funcB) that could return `imageDigestMirror` object

rccrdpccl:
please add relevant documentation section

controlplane/internal/controller/imageregistry/image_registry.go
CrystalChun: As of right now, it skips adding the registry (in case the user wants to just provide an additional certificate and not a registry), but do you think we should not allow that case? Or we should bubble up a warning?

controlplane/internal/controller/imageregistry/image_registry.go
CrystalChun: Good catch, thanks!

controlplane/internal/controller/imageregistry/image_registry.go
rccrdpccl: Can you please expand on this use case? Why would the user just add a certificate and not a registry?

Asking because we also have AdditionalCertificateBundle - wouldn't this be the proper way to just add certs?

CrystalChun: Ah yes, the proper way to add additional certificates is through that field. Thanks!! So here we could return an error if the registry is not found, what do you think?

controlplane/internal/controller/imageregistry/image_registry.go
```diff
@@ -0,0 +1,199 @@
+package imageRegistry
```
rccrdpccl, line 1: probably this should not be under controller, but at the same level.
also package names should not be camelcase, rather `imageregistry`

rccrdpccl: to clarify, not `controlplane/internal/controller` but `controlplane/internal/imageregistry`

CrystalChun: Ah gotcha, thanks for clarifying! Changed it in this PR https://github.com/openshift-assisted/cluster-api-agent/pull/24

controlplane/internal/controller/imageregistry/image_registry.go
CrystalChun: Modified in this PR https://github.com/openshift-assisted/cluster-api-agent/pull/24

controlplane/internal/controller/imageregistry/image_registry.go
CrystalChun: Thanks!! Modified here https://github.com/openshift-assisted/cluster-api-agent/pull/24

PR #21, opened 2024-06-12
fix tests in bootstrap provider

rccrdpccl:


bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -418,14 +416,7 @@ func (r *AgentBootstrapConfigReconciler) handleDeletion(ctx context.Context, con
 			}
 			config.Status.AgentRef = nil
 		}
-
-		// Remove finalizer to delete
-		if controllerutil.RemoveFinalizer(config, agentBootstrapConfigFinalizer) {
```
rccrdpccl, line 423: This function will modify the `config` object, which will be patched with the deferred function

PR #27, opened 2024-06-14
fix tests in controlplane controller

rccrdpccl:


controlplane/internal/controller/agentcontrolplane_controller_test.go
```diff
@@ -127,45 +110,28 @@ var _ = Describe("AgentControlPlane Controller", func() {
...
+			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{NamespacedName: typeNamespacedName})
+			Expect(err).NotTo(HaveOccurred())
 
 			By("checking if the agent control plane has the finalizer")
 			err = k8sClient.Get(ctx, typeNamespacedName, agentControlPlane)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(agentControlPlane.Finalizers).NotTo(BeEmpty())
 			Expect(agentControlPlane.Finalizers).To(ContainElement(acpFinalizer))
 		})
-
```
rccrdpccl, line 143: Unfortunately this type of test has several issues with fake client, around deletion and how long this object is present after deletion (cannot verify finalizer is being removed).

I'd suggest let's run these kind of tests with EnvTest, in a later stage

PR #29, opened 2024-06-14
MGMT-18017: Create placeholder pull secret

CrystalChun:
Assisted installer requires a pull secret for every cluster created, however, OKD does not need any authentication to pull its images. This creates a placeholder pull secret in the event that one isn't needed or provided.

### DIFF
--- a/controlplane/internal/controller/agentcontrolplane_controller.go
+++ b/controlplane/internal/controller/agentcontrolplane_controller.go
@@ -18,12 +18,13 @@ package controller
 
 import (
 	"context"
+	"time"
+
 	"github.com/openshift-assisted/cluster-api-agent/assistedinstaller"
 	"github.com/openshift-assisted/cluster-api-agent/util"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/reference"
 	"sigs.k8s.io/cluster-api/util/patch"
-	"time"
 
 	bootstrapv1alpha1 "github.com/openshift-assisted/cluster-api-agent/bootstrap/api/v1alpha1"
 	logutil "github.com/openshift-assisted/cluster-api-agent/util/log"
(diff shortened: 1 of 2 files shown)


CrystalChun:
Merging, feel free to comment and will change in future.

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -452,3 +450,39 @@ func (r *AgentControlPlaneReconciler) generateAgentBootstrapConfig(ctx context.C
...
+	}
+	acp.Spec.AgentConfigSpec.PullSecretRef = &corev1.LocalObjectReference{Name: secret.Name}
+	return r.Client.Update(ctx, acp)
+}
+
+// Assisted-service expects the pull secret to
+// 1. Have .dockerconfigjson as a key
+// 2. Have the value of .dockerconfigjson be a base64-encoded JSON
+// 3. The JSON must have the key "auths" followed by repository with an "auth" key
+func GenerateSecret(name, namespace string) *corev1.Secret {
```
rccrdpccl, line 475: Does this need to be in `utils`? Is it going to be used anywhere else? If not, we can create a new file in the package where it's used, i.e. `auth.go`

rccrdpccl:
Can we make sure we have tests cases covering both branches (existing pull secret and no pull secret, where we'd need to generate it)?

controlplane/internal/controller/agentcontrolplane_controller.go
CrystalChun: Moved here https://github.com/openshift-assisted/cluster-api-agent/pull/43

CrystalChun:
> Can we make sure we have tests cases covering both branches (existing pull secret and no pull secret, where we'd need to generate it)?

Yep! Added test case with no pull secret here https://github.com/openshift-assisted/cluster-api-agent/pull/43 and the one with existing pull secret existed already https://github.com/openshift-assisted/cluster-api-agent/pull/43/files#diff-76355a50b21e79551ce9096642c98c7cd4b56c590e2ccfba76300048d623d943L159

PR #38, opened 2024-06-21
MGMT-17963: Refactor image registry

CrystalChun:
Limit scope of image registry file to creating the configmap for the image registry instead of accessing the kube api server.

Also moves logic of creating the image registry configmap to the cluster deployment controller and includes an update of the image registry unit tests to reflect the move.

### DIFF
--- a/controlplane/internal/controller/clusterdeployment_controller.go
+++ b/controlplane/internal/controller/clusterdeployment_controller.go
@@ -276,18 +276,21 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
 	}
 
 	if acp.Spec.AgentConfigSpec.ImageRegistryRef != nil {
-		imageRegistryManifest, err := imageregistry.CreateConfig(
-			ctx,
-			r.Client,
-			acp.Spec.AgentConfigSpec.ImageRegistryRef,
-			clusterDeployment.Namespace,
-		)
+		registryConfigmap := &corev1.ConfigMap{}
+		if err := r.Client.Get(ctx, client.ObjectKey{Name: acp.Spec.AgentConfigSpec.ImageRegistryRef.Name, Namespace: acp.Namespace}, registryConfigmap); err != nil {
+			return nil, err
+		}
+
+		spokeImageRegistryConfigmap, err := imageregistry.CreateImageRegistryConfigmap(registryConfigmap, acp.Namespace)
 		if err != nil {
+			return nil, err
+		}
+
+		if _, err := ctrl.CreateOrUpdate(ctx, r.Client, spokeImageRegistryConfigmap, func() error { return nil }); err != nil && !apierrors.IsAlreadyExists(err) {
 			log.Error(err, "failed to create image registry config manifest")
 			return nil, err
-		} else {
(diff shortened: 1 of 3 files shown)


CrystalChun:
Merging for now to release, feel free to comment and will update in future!

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -276,18 +276,21 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
...
+			return nil, err
+		}
+
+		if _, err := ctrl.CreateOrUpdate(ctx, r.Client, spokeImageRegistryConfigmap, func() error { return nil }); err != nil && !apierrors.IsAlreadyExists(err) {
 			log.Error(err, "failed to create image registry config manifest")
 			return nil, err
-		} else {
-			additionalManifests = append(additionalManifests, hiveext.ManifestsConfigMapReference{Name: imageRegistryManifest})
 		}
+		additionalManifests = append(additionalManifests, hiveext.ManifestsConfigMapReference{Name: spokeImageRegistryConfigmap.Name})
```
rccrdpccl, line 293: I think this can be extracted to something like:

```golang
if acp.Spec.AgentConfigSpec.ImageRegistryRef != nil {
    if cm, err := r.createImageRegistryConfigmap(...); err != nil {
          additionalManifests = append(additionalManifests, hiveext.ManifestsConfigMapReference{Name: cm.Name})
    }
}
```

controlplane/internal/imageregistry/image_registry.go
```diff
@@ -28,35 +23,12 @@ const (
...
-	return additionalRegistryConfigMap.Name, nil
-}
-
-func createImageRegistryConfig(imageRegistry *corev1.ConfigMap, namespace string) (*corev1.ConfigMap, error) {
+// CreateImageRegistryConfigmap creates a ConfigMap containing the manifests for mirror registry configuration
+// to be created on the spoke cluster. The manifests that are in the ConfigMap include an ImageDigestMirrorSet CR
+// and/or an ImageTagMirrorSet CR with the mirror registry information, a ConfigMap containing the additional
+// trusted certificates for the mirror registry, and an image.config.openshift.io CR that references the
+// ConfigMap with the additional trusted certificate.
+func CreateImageRegistryConfigmap(imageRegistry *corev1.ConfigMap, namespace string) (*corev1.ConfigMap, error) {
```
rccrdpccl, line 31: I think "Create" could be an ambiguous term. Is not really creating the resource, right? It will generate the right struct, that later we can create. I'd suggest to change it to something less ambiguous, as `NewImageRegistryConfigmap` or `generateImageRegistryConfigmap` or something else.

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Great idea, thanks!! Modified in https://github.com/openshift-assisted/cluster-api-agent/pull/39

controlplane/internal/imageregistry/image_registry.go
CrystalChun: Agreed, thanks for pointing this out! Modified in https://github.com/openshift-assisted/cluster-api-agent/pull/39

PR #39, opened 2024-06-24
MGMT-17963: Return clearer error messages for image registry

CrystalChun:
Creates and returns error messages that indicate where the image registry creation failed.

Includes small refactoring of creating the registry configmap and test cleanup.

### DIFF
--- a/controlplane/internal/imageregistry/image_registry.go
+++ b/controlplane/internal/imageregistry/image_registry.go
@@ -6,6 +6,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pelletier/go-toml"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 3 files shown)


rccrdpccl:
sorry for causing the conflict @CrystalChun , just wanted to make sure we add linter to the process

CrystalChun:
> sorry for causing the conflict @CrystalChun , just wanted to make sure we add linter to the process

No problem! What do you think of the error messages?

controlplane/internal/imageregistry/image_registry_test.go
```diff
@@ -217,13 +226,13 @@ var _ = Describe("ImageRegistry Test", func() {
...
-				imageRegistryConfigMap, err := CreateImageRegistryConfigmap(
+				By("Calling the GenerateImageRegistryConfigmap function")
+				imageRegistryConfigMap, err := GenerateImageRegistryConfigmap(
 					userRegistryCM,
-					namespace,
+					testNamespace,
 				)
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(Equal("(6, 13): cannot have two dots in one float"))
+				Expect(err.Error()).To(ContainSubstring("(6, 13): cannot have two dots in one float"))
```
rccrdpccl, line 235: Can we see an example of descriptive error message here? Would this be `failed to parse image mirror registry toml for registry <name> : (6, 13): cannot have two dots in one float`?

controlplane/internal/imageregistry/image_registry.go
```diff
@@ -142,28 +148,30 @@ func parseMirrorRegistries(
...
+				itmsMirror.Source = source
+				itmsMirror.Mirrors = append(itmsMirror.Mirrors, configv1.ImageMirror(mirror))
+			case "digest-only":
 				idmsMirror.Source = source
 				idmsMirror.Mirrors = append(idmsMirror.Mirrors, configv1.ImageMirror(mirror))
 			}
+		} else {
+			// Default is pulling by digest
+			idmsMirror.Source = source
+			idmsMirror.Mirrors = append(idmsMirror.Mirrors, configv1.ImageMirror(mirror))
```
rccrdpccl, line 174: Can we extract this to a function? Doing so we'll also get rid of the `else` condition.

CrystalChun: Modified, how does it look?

controlplane/internal/imageregistry/image_registry_test.go
CrystalChun: Yeah! Edited to match the error message

controlplane/internal/imageregistry/image_registry_test.go
```diff
@@ -217,13 +226,14 @@ var _ = Describe("ImageRegistry Test", func() {
...
+				By("Calling the GenerateImageRegistryConfigmap function")
+				imageRegistryConfigMap, err := GenerateImageRegistryConfigmap(
 					userRegistryCM,
-					namespace,
+					testNamespace,
 				)
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(Equal("(6, 13): cannot have two dots in one float"))
+				Expect(err.Error()).To(Equal(
+					"failed to load value of registries.conf into toml tree; incorrectly formatted toml: (6, 13): cannot have two dots in one float"))
```
rccrdpccl, line 236: much clearer :)

CrystalChun:
Will be included in https://github.com/openshift-assisted/cluster-api-agent/pull/41

PR #41, opened 2024-06-25
MGMT-17983: Refactor controlplane/clusterdeployment controller

CrystalChun:
Includes changes from this PR https://github.com/openshift-assisted/cluster-api-agent/pull/39

### DIFF
--- a/controlplane/internal/controller/clusterdeployment_controller.go
+++ b/controlplane/internal/controller/clusterdeployment_controller.go
@@ -35,7 +35,6 @@ import (
 	capiutil "sigs.k8s.io/cluster-api/util"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 )
 
@@ -209,23 +208,11 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
 	}
 
 	if acp.Spec.AgentConfigSpec.ImageRegistryRef != nil {
-		registryConfigmap := &corev1.ConfigMap{}
-		if err := r.Client.Get(ctx, client.ObjectKey{Name: acp.Spec.AgentConfigSpec.ImageRegistryRef.Name, Namespace: acp.Namespace}, registryConfigmap); err != nil {
-			return nil, err
-		}
-
-		spokeImageRegistryConfigmap, err := imageregistry.CreateImageRegistryConfigmap(registryConfigmap, acp.Namespace)
-		if err != nil {
-			return nil, err
-		}
-
-		_ = controllerutil.SetOwnerReference(&acp, spokeImageRegistryConfigmap, r.Scheme)
-
-		if _, err := ctrl.CreateOrUpdate(ctx, r.Client, spokeImageRegistryConfigmap, func() error { return nil }); err != nil && !apierrors.IsAlreadyExists(err) {
+		if err := r.createImageRegistry(ctx, acp.Spec.AgentConfigSpec.ImageRegistryRef.Name, acp.Namespace); err != nil {
 			log.Error(err, "failed to create image registry config manifest")
 			return nil, err
 		}
-		additionalManifests = append(additionalManifests, hiveext.ManifestsConfigMapReference{Name: spokeImageRegistryConfigmap.Name})
+		additionalManifests = append(additionalManifests, hiveext.ManifestsConfigMapReference{Name: imageregistry.ImageConfigMapName})
 	}
 
 	aci := &hiveext.AgentClusterInstall{
@@ -260,4 +247,21 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
 		},
 	}
 	return aci, nil
+}
+
+func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, registryName, registryNamespace string) error {
+	registryConfigmap := &corev1.ConfigMap{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: registryName, Namespace: registryNamespace}, registryConfigmap); err != nil {
+		return err
+	}
+
+	spokeImageRegistryConfigmap, err := imageregistry.GenerateImageRegistryConfigmap(registryConfigmap, registryNamespace)
+	if err != nil {
+		return err
+	}
+
+	if _, err := ctrl.CreateOrUpdate(ctx, r.Client, spokeImageRegistryConfigmap, func() error { return nil }); err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
 }
--- a/controlplane/internal/controller/clusterdeployment_controller.go
+++ b/controlplane/internal/controller/clusterdeployment_controller.go
@@ -18,12 +18,11 @@ package controller
 
 import (
 	"context"
-	"time"
 
-	"github.com/openshift-assisted/cluster-api-agent/util"
-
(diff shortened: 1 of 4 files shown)


controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -67,42 +68,30 @@ func (r *ClusterDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		return ctrl.Result{}, nil
 	}
 
-	for _, acp := range agentCPList.Items {
-		if IsAgentControlPlaneReferencingClusterDeployment(acp, clusterDeployment) {
-			log.Info("ClusterDeployment is referenced by AgentControlPlane")
-			return r.ensureAgentClusterInstall(ctx, clusterDeployment, acp)
-		}
+	acp := controlplanev1alpha1.AgentControlPlane{}
+	if err := util.GetTypedOwner(ctx, r.Client, clusterDeployment, &acp); err != nil {
```
rccrdpccl, line 80: Wouldn't this cover also the previous condition? Would we care enough to have a specific condition if not enough references?

CrystalChun: The GetTypedOwner returns an error if there's no owners, which would cause this ClusterDeployment to be reconciled again. 

I think we should ignore any ClusterDeployment (not reconcile again) if there are no owners. What do you think?

rccrdpccl: What's the difference between "there is no owner" and "there are owners, just not of the type you are looking for"? Wouldn't we need to reconcile in the same way? (Either requeue or not requeue on both?)

CrystalChun: That's true. What I want to avoid is requeueing a clusterdeployment that doesn't belong to us (cluster api agent). I'll change this to reflect that

rccrdpccl:
I think it would've been better to keep https://github.com/openshift-assisted/cluster-api-agent/pull/39 separate from this PR, so we don't mix different changes in the same PR

controlplane/internal/imageregistry/image_registry.go
```diff
@@ -112,10 +113,15 @@ func getImageRegistries(registry string) (string, string, []string, error) {
...
-			continue
+		if !sourceExists {
+			return "", "", nil, fmt.Errorf("failed to parse image registry toml, missing registry location")
+		}
+		if !mirrorExists {
+			return "", "", nil, fmt.Errorf("failed to parse image registry toml, missing registry.mirror in registries.conf")
+		}
+		if err := parseMirrorRegistries(&idmsMirrors, &itmsMirrors, mirrorTrees, &insecureRegistries, source); err != nil {
+			return "", "", nil, errors.Wrap(err, fmt.Sprintf("failed to parse image mirror registry toml for registry %s", source))
 		}
```
rccrdpccl, line 124: I think we can reorganize the `getImageRegistries` function, maybe using a specific struct to encapsulate values from current `parse` function (that will need renaming). For example:
```golang
type RegistryConfig struct {
    TagMirrors *configv1.ImageTagMirrors{}
    DigestMirrors *configv1.ImageDigestMirrors{}
    InsecureRegistries []string
}
...

configs := []RegistryConfig{}
for _, registry := range registriesTree {
	source, sourceExists := registry.Get("location").(string)
	mirrorTrees, mirrorExists := registry.Get("mirror").([]*toml.Tree)
        // handle errors
        if registryConfig, err := getRegistryConfig(source, mirrorTrees); err != nil {
        // handle errors
        }
        configs = append(configs, registryConfig)
}

...
generateImageDigestMirrorSet(configs)
generateImageTagMirrorSet(configs)
generateInsecureRegistries(configs)
generateOpenshiftImageConfig(configs, registryCertExists)
```

Or something along these lines. WDYT?

controlplane/internal/imageregistry/image_registry.go
```diff
@@ -142,36 +148,43 @@ func parseMirrorRegistries(
...
-				idmsMirror.Mirrors = append(idmsMirror.Mirrors, configv1.ImageMirror(mirror))
-			}
+		mirror, ok := mirrorTree.Get("location").(string)
+		if !ok {
+			return fmt.Errorf("missing mirror registry location key or the value of location is not a string")
+		}
+		if insecure, ok := mirrorTree.Get("insecure").(bool); ok && insecure {
+			*insecureRegistries = append(*insecureRegistries, mirror)
 		}
+		setPullPolicy(itmsMirror, idmsMirror, source, mirror, mirrorTree)
```
rccrdpccl, line 162: we can already know that the mirror is going to be `configv1.ImageMirror(mirror)`, also that itmsMirror.Source and idmsMirror.Source is `source`. We will then need to decide where to add the mirror: will it be itms or idms? 

```golang
mirror, isInsecure := getMirrorFromToml(mirrorTree)
imageMirror := configv1.ImageMirror(mirror)
if pullFrom, ok := mirrorTree.Get("pull-from-mirror").(string); ok && pullFrom == "tag-only" {
    itmsMirrors.Mirrors = append(itmsMirrors.Mirrors, imageMirror)
    continue
}
idmsMirrors.Mirrors = append(idmsMirrors.Mirrors, imageMirror)
```

Also I think we can reorganize the `getImageRegistries` function, maybe using a specific struct to encapsulate values from current `parse` function (that will need renaming). For example:
```golang
type RegistryConfig struct {
    TagMirrors *configv1.ImageTagMirrors{}
    DigestMirrors *configv1.ImageDigestMirrors{}
    InsecureRegistries []string
}

File: controlplane/internal/imageregistry/image_registry.go (hunk 2 above)
Comment (CrystalChun) [reply controlplane/internal/imageregistry/image_registry.go]: Good idea! Modified it, how does it look?

File: controlplane/internal/imageregistry/image_registry.go (hunk 3 above)
Comment (CrystalChun) [reply controlplane/internal/imageregistry/image_registry.go]: Good idea! Modified it, how does it look?

File: controlplane/internal/imageregistry/image_registry.go [hunk 4]
```diff
@@ -93,91 +119,73 @@ func CreateImageRegistryConfigmap(imageRegistry *corev1.ConfigMap, namespace str
...
 		mirrorTrees, mirrorExists := registry.Get("mirror").([]*toml.Tree)
-		if !sourceExists || !mirrorExists {
-			continue
+		if !sourceExists {
+			return fmt.Errorf("failed to parse image registry toml, missing registry location")
+		}
+		if !mirrorExists {
+			return fmt.Errorf("failed to parse image registry toml, missing registry.mirror in registries.conf")
+		}
+		if err := parseMirrorRegistries(mirrorConfigs, mirrorTrees, source); err != nil {
```
rccrdpccl, line 143: Why not returning only an object here with ImageTagMirrors, ImageDigestMirrors and insecure registry, and move the merge/addition logic outside?
The benefit I see with this is that we would use a more functional approach, no need to provide any data to the function and it will always return the same value with the same input, no matter what we have been parsing before.

controlplane/internal/imageregistry/image_registry.go
```diff
@@ -41,14 +56,25 @@ func CreateImageRegistryConfigmap(imageRegistry *corev1.ConfigMap, namespace str
 		)
 	}
 
-	imageDigestMirrors, imageTagMirrors, insecureRegistries, err := getImageRegistries(registryConf)
+	mirrorConfigs := initializeMirrorConfigs()
```
rccrdpccl, line 59: No need to initialize explicitly, we can do it inside getImageRegistry

controlplane/internal/imageregistry/image_registry.go
CrystalChun: Ok modified it!

controlplane/internal/imageregistry/image_registry.go
CrystalChun: Good idea, modified it!

PR #42, opened 2024-06-25
fix linter and add unit test coverage

rccrdpccl:


controlplane/internal/imageregistry/image_registry_test.go
```diff
@@ -232,57 +202,108 @@ var _ = Describe("ImageRegistry Test", func() {
...
-		registryToml = fmt.Sprintf("%s\n[[registry.mirror]]\nlocation = \"%s\"", registryToml, mirror)
-	}
-	if insecure {
-		registryToml = fmt.Sprintf("%s\ninsecure = true", registryToml)
-	}
-	if pullByTag {
-		registryToml = fmt.Sprintf("%s\npull-from-mirror = \"tag-only\"", registryToml)
-	}
-	return registryToml
+func getSecureRegistryToml() string {
```
rccrdpccl, line 234: @CrystalChun I think trying to use code as simple as possible in the tests always helps. Getting rid of conditions makes clearer what we are testing, and reduce probability of errors in the test itself, otherwise we'd need to write a test for the test :)

PR #43, opened 2024-06-25
MGMT-18017: Refactor placeholder pull secret

CrystalChun:
Moving the function to its own auth file as it's only used by the Agent Control Plane controller.

### DIFF
--- a/controlplane/internal/auth/auth.go
+++ b/controlplane/internal/auth/auth.go
@@ -1,0 +1,25 @@
+package auth
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Assisted-service expects the pull secret to
+// 1. Have .dockerconfigjson as a key
+// 2. Have the value of .dockerconfigjson be a base64-encoded JSON
+// 3. The JSON must have the key "auths" followed by repository with an "auth" key
+func GenerateFakePullSecret(name, namespace string) *corev1.Secret {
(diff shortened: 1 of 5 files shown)


controlplane/internal/auth/auth.go
```diff
@@ -0,0 +1,25 @@
+package auth
```
rccrdpccl, line 1: I was referring to the same `controller` package but I guess this would work as well

controlplane/internal/controller/agentcontrolplane_controller_test.go
```diff
@@ -159,6 +159,46 @@ var _ = Describe("AgentControlPlane Controller", func() {
 				Expect(cd.Spec.PullSecretRef).To(Equal(agentControlPlane.Spec.AgentConfigSpec.PullSecretRef))
 			})
 		})
+
+		When("a pull secret isn't set on the AgentControlPlane", func() {
+			It("should successfully create the ClusterDeployment with the default fake pull secret", func() {
+				By("setting the cluster as the owner ref on the agent control plane")
+				agentControlPlane := getAgentControlPlane()
+				agentControlPlane.SetOwnerReferences(
```
rccrdpccl, line 167: Linter will complain about return value

rccrdpccl: Nevermind, I got confused with the other utility method, sorry :sweat_smile:

controlplane/internal/auth/auth.go
CrystalChun: Oh oops, do you have a preference?

PR #50, opened 2024-07-19
Baremetal install

rccrdpccl:
Add multinode baremetal install capability


bootstrap/config/rbac/role.yaml
```diff
@@ -109,24 +68,18 @@ rules:
 - apiGroups:
   - cluster.x-k8s.io
   resources:
-  - machines
-  - machines/status
+  - machinesets
+  - machinesets/status
```
CrystalChun, line 72: Note for myself to add this to RBAC PR

bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -443,20 +450,27 @@ func (r *AgentBootstrapConfigReconciler) getInfrastructureRefKey(
 	machine *clusterv1.Machine,
 ) (types.NamespacedName, error) {
 	acp := controlplanev1alpha1.AgentControlPlane{}
+	namespace := acp.Namespace
 	err := util.GetTypedOwner(ctx, r.Client, machine, &acp)
 	if err != nil {
-		// Machine is not owned by ACP, check for MD
-		md := clusterv1.MachineDeployment{}
+		// Machine is not owned by ACP, check for MS
+		md := clusterv1.MachineSet{}
```
CrystalChun, line 457: Super minor nit but if we changed this to MachineSet, maybe the var name could match it? 
```suggestion
		ms := clusterv1.MachineSet{}
```

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -327,5 +328,14 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
 			ManifestsConfigMapRefs: additionalManifests,
 		},
 	}
+
+	if len(acp.Spec.AgentConfigSpec.APIVIPs) > 0 && len(acp.Spec.AgentConfigSpec.IngressVIPs) > 0 {
+		aci.Spec.APIVIPs = acp.Spec.AgentConfigSpec.APIVIPs
+		aci.Spec.IngressVIPs = acp.Spec.AgentConfigSpec.IngressVIPs
```
CrystalChun, line 334: What version of assisted-service will our operator support?

I ask because previous version require `aci.Spec.APIVIP` and I'm wondering if we should have that too?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -327,5 +328,14 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
...
+		aci.Spec.PlatformType = hiveext.PlatformType(configv1.BareMetalPlatformType)
```
CrystalChun, line 335: Should we also expose `PlatformType` field in our AgentControlPlane spec?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -327,5 +328,14 @@ func (r *ClusterDeploymentReconciler) computeAgentClusterInstall(
 			ManifestsConfigMapRefs: additionalManifests,
 		},
 	}
+
+	if len(acp.Spec.AgentConfigSpec.APIVIPs) > 0 && len(acp.Spec.AgentConfigSpec.IngressVIPs) > 0 {
```
CrystalChun, line 332: Also should we have a check that if it's a multinode cluster, there should be VIPs set?

bootstrap/config/rbac/role.yaml
CrystalChun: Also we don't seem to look for machine deployments anymore, so could remove that

controlplane/internal/controller/clusterdeployment_controller.go
rccrdpccl: I don't think so, it could be UMN, right?

controlplane/internal/controller/clusterdeployment_controller.go
rccrdpccl: I feel like this detail is not really necessary. Do you have any usecase in mind where this would help?

controlplane/internal/controller/clusterdeployment_controller.go
rccrdpccl: I think this was quite some versions ago, but we can check and make sure, and either:
* before such version, just set APIVIP instead (maybe just pick the first item to keep our API simpler)
* Do not allow installs of what it's probably ancient versions
WDYT?

CrystalChun: The second option is more favorable imo, since we don't want to extend support to prior versions. Do you think it'll be difficult to detect which version is deployed?

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Just wondering in the event the user might want platformType none since that's still a viable option right? Unless we're really only limiting the scope to Sylva and they only want platformType baremetal

rccrdpccl: Yes, platform none is still available. It will be used if we do not specify VIPs (and for SNO)

CrystalChun: Gotcha! Thanks for confirming!

controlplane/internal/controller/clusterdeployment_controller.go
rccrdpccl: I have added naive checking vs the string version, I think we could introspect the image and extract `io.openshift.release` or something to make it better, I will create an issue for that.

bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -483,9 +496,14 @@ func (r *AgentBootstrapConfigReconciler) setMetal3MachineTemplateImage(
 
 	if machineTpl.Spec.Template.Spec.Image.URL != config.Status.ISODownloadURL ||
 		machineTpl.Spec.Template.Spec.Image.DiskFormat != &liveIsoFormat {
+		patchHelper, err := patch.NewHelper(machineTpl, r.Client)
 		machineTpl.Spec.Template.Spec.Image.URL = config.Status.ISODownloadURL
 		machineTpl.Spec.Template.Spec.Image.DiskFormat = &liveIsoFormat
-		if err := r.Client.Update(ctx, machineTpl); err != nil {
+		if err != nil {
+			log.Error(err, "couldn't create patch helper for machineTpl")
```
CrystalChun, line 503: Should we do something here? If there's an error creating the patchHelper then we probably shouldn't run this line right? https://github.com/openshift-assisted/cluster-api-agent/pull/50/files#diff-f7f355380e3be4093bf66be019c03d3592f6d6b366191d2fc800a57eb1599241R506

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -123,6 +128,17 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		controllerutil.AddFinalizer(acp, acpFinalizer)
 	}
 
+	acpVersion, err := semver.NewVersion(acp.Spec.Version)
+	if err != nil {
+		log.Error(err, "invalid version", "version", acp.Spec.Version)
```
CrystalChun, line 133:
```suggestion
		log.Error(err, "invalid version openshift", "version", acp.Spec.Version)
```
Minor: Maybe a good idea to let them know the version of what?

controlplane/internal/controller/agentcontrolplane_controller.go
```diff
@@ -123,6 +128,17 @@ func (r *AgentControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		return ctrl.Result{}, err
+	}
+	if acpVersion.LessThan(*minVersion) {
+		conditions.MarkFalse(acp, controlplanev1alpha1.MachinesCreatedCondition, controlplanev1alpha1.MachineGenerationFailedReason,
+			clusterv1.ConditionSeverityError, fmt.Sprintf("version %v is not supported", acp.Spec.Version))
```
CrystalChun, line 138:
```suggestion
			clusterv1.ConditionSeverityError, fmt.Sprintf("openshift version %v is not supported, must be greater than 4.14.0", acp.Spec.Version))
```

Minor: maybe a good idea to let them know the version of what and what's expected?

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Nice! Yeah sounds good

PR #51, opened 2024-07-19
MGMT-18418: Use Image Service HTTP IP for live iso URL

CrystalChun:
Ironic is unable to pull using assisted-service's CA certificate from changes in https://github.com/openshift/assisted-service/pull/6564 https://issues.redhat.com/browse/MGMT-18418

Adds the option to use the internal IP of the service as the URL to provision the BMH so Ironic can pull the image without TLS verification.


bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -18,23 +18,40 @@ package controller
...
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type InfraEnvControllerConfig struct {
+	// UseInsecureImageURL when set to false means that we'll use the InfraEnv's iso download URL
+	// as is. When set to true, it'll find the assisted-image-service's internal IP as part of the
+	// download URL.
+	UseInsecureImageURL bool `envconfig:"USE_INSECURE_IMAGE_URL" default:"false"`
```
carbonin, line 43: Maybe "Internal" instead of "Insecure"?

bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -77,10 +101,30 @@ func (r *InfraEnvReconciler) attachISOToAgentBootstrapConfigs(ctx context.Contex
...
+		return "", errors.Wrap(err, "failed to find assisted image service service")
+	}
+
+	_, remainderURL, found := strings.Cut(originalURL, "byapikey") //TODO: will the URL always have this?
+	if !found {
+		return "", fmt.Errorf("failed to parse InfraEnv Download URL %s", originalURL)
+	}
+
+	//TODO: better way to build this string
+	downloadURL := "http://" + svc.Spec.ClusterIP + ":" + strconv.Itoa(int(svc.Spec.Ports[0].Port)) + "/byapikey" + remainderURL
```
carbonin, line 128: You should do all of this with the standard library [net/url](https://pkg.go.dev/net/url@go1.22.5)

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: That sounds much better, thanks!

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: Thank you!!

CrystalChun: Modified to use the std lib net/url package!

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: Modified!

bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -18,23 +18,40 @@ package controller
...
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type InfraEnvControllerConfig struct {
+	// UseInternalImageURL when set to false means that we'll use the InfraEnv's iso download URL
+	// as is. When set to true, it'll use the assisted-image-service's internal IP as part of the
+	// download URL.
+	UseInternalImageURL bool `envconfig:"USE_INTERNAL_IMAGE_URL" default:"false"`
```
carbonin, line 43: Should we set this to true by default?

bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -18,23 +18,40 @@ package controller
...
+	// ImageServiceName is the Service CR name for the assisted-image-service
+	ImageServiceName string `envconfig:"IMAGE_SERVICE_NAME" default:"assisted-image-service"`
+	// ImageServiceNamespace is the namespace that the Service CR for the assisted-image-service is in
+	ImageServiceNamespace string `envconfig:"IMAGE_SERVICE_NAMESPACE" default:"assisted-installer"`
```
carbonin, line 47: Do we have a good idea of how this will be set on deployment?

bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -77,10 +101,34 @@ func (r *InfraEnvReconciler) attachISOToAgentBootstrapConfigs(ctx context.Contex
...
+		return "", fmt.Errorf("failed to get internal image service URL, either cluster IP or Ports were missing from Service")
+	}
+
+	isoURL, err := url.Parse(originalURL)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to parse InfraEnv ISO download URL %s", originalURL)
+	}
+
+	isoURL.Scheme = "http"
+	isoURL.Host = fmt.Sprintf("%s:%s", svc.Spec.ClusterIP, strconv.Itoa(int(svc.Spec.Ports[0].Port)))
```
carbonin, line 132: You can probably use `%d` here instead of strconv and a cast.
```suggestion
	isoURL.Host = fmt.Sprintf("%s:%d", svc.Spec.ClusterIP, svc.Spec.Ports[0].Port)
```

CrystalChun: Ah good point 🤦‍♀️ thank you!! Modified

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: Sure, modified to be true

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: Probably as part of the `env` field either set directly or part of a configmap. @rccrdpccl?

bootstrap/internal/controller/infraenv_controller_test.go
```diff
@@ -156,4 +160,169 @@ var _ = Describe("InfraEnv Controller", func() {
...
+				WithStatusSubresource(&bootstrapv1alpha1.AgentBootstrapConfig{}).
+				Build()
+			Expect(k8sClient).NotTo(BeNil())
+
+			controllerReconciler = &InfraEnvReconciler{
+				Client: k8sClient,
+				Scheme: k8sClient.Scheme(),
+				Config: InfraEnvControllerConfig{
+					UseInternalImageURL:   true,
+					ImageServiceNamespace: namespace,
```
rccrdpccl, line 180: Can we change it to simulate a more realistic scenario and change it to a different namespace?

bootstrap/config/manager/manager.yaml
```diff
@@ -73,6 +73,13 @@ spec:
         image: controller:latest
         imagePullPolicy: Always
         name: manager
+        env:
+        - name: USE_INTERNAL_IMAGE_URL
+          value: "true"
+        - name: IMAGE_SERVICE_NAME
+          value: "assisted-image-service"
+        - name: IMAGE_SERVICE_NAMESPACE
+          value: "assisted-installer"
```
carbonin, line 82: Do we want to allow this to be configurable through the kustomize files somehow?

rccrdpccl: I think we do. We should make it easy to customize and maybe default to false. @jianzzha do you have any suggestion about this?

CrystalChun: We can also set the default here too https://github.com/openshift-assisted/cluster-api-agent/pull/51#discussion_r1687128686

jianzzha: I think this is ok.

To kustomize an env var like this, a hard coded index has to be used in the kustomization file. Say to kustomize  USE_INTERNAL_IMAGE_URL, in the kustomization file, index 0 will be used below
```- op: replace
        path: /spec/template/spec/containers/0/env/0/value
```

To make sure the kustomization file does not break when new env vars are added later,  new env vars have to be appended, not insert before the existing one - something to remember :)

CrystalChun: Great to know, thank you for your input!!

bootstrap/internal/controller/infraenv_controller_test.go
CrystalChun: Changed it to a different namespace, thanks!!

bootstrap/config/manager/manager.yaml
CrystalChun: I added an `overlays/ocp` folder to do a kustomize patch - what do you all think?

bootstrap/config/manager/overlays/ocp/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+patches:
+  - target:
+      kind: Deployment
+      name: controller-manager
+    patch: |-
+      - op: replace
+        path: "/spec/template/spec/containers/0/env/0"
+        value:
+          name: USE_INTERNAL_IMAGE_URL
```
carbonin, line 15: I thought this was meant to allow someone deploying this to change the namespace for assisted. How does this do that?

carbonin (commented):
@jianzzha can you put a review here? I want to make sure whatever we're building will work for how you want to deploy this.

bootstrap/config/manager/overlays/ocp/kustomization.yaml
CrystalChun: This stanza replaces the first environment variable listed in the base deployment yaml with the `value` specified here. 

So the result of `$ kustomize build bootstrap/config/manager/overlays/ocp`
is this:
```yaml
...
apiVersion: apps/v1
kind: Deployment
metadata:
...
  name: controller-manager
  namespace: system
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/default-container: manager
      labels:
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --leader-elect
        command:
        - /manager
        env:
        - name: USE_INTERNAL_IMAGE_URL
          value: "false"
        - name: IMAGE_SERVICE_NAME
          value: assisted-image-service
        - name: IMAGE_SERVICE_NAMESPACE
          value: assisted-installer
        image: quay.io/edge-infrastructure/openshift-capi-agent-bootstrap:latest
        imagePullPolicy: Always
        livenessProbe:
...
```

Whereas just doing `$ kustomize build bootstrap/config/manager/base`
produces this:
```yaml      
...
apiVersion: apps/v1
kind: Deployment
metadata:
...
  name: controller-manager
  namespace: system
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/default-container: manager
      labels:
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --leader-elect
        command:
        - /manager
        env:
        - name: USE_INTERNAL_IMAGE_URL
          value: "true"
        - name: IMAGE_SERVICE_NAME
          value: assisted-image-service
        - name: IMAGE_SERVICE_NAMESPACE
          value: assisted-installer
        image: quay.io/edge-infrastructure/openshift-capi-agent-bootstrap:latest
        imagePullPolicy: Always
        livenessProbe:
...
```

jianzzha: For the purpose of the sylva integration, the released manifests are used. For example if release 0.1.5 will have this fix, it's the github.com/openshift-assisted/cluster-api-agent/releases/download/v0.1.5/{bootstrap-components.yaml,controlplane-components.yaml} will be used by Sylva. Since bootstrap-components.yaml is already updated, so I think we are good. We can always overwrite  USE_INTERNAL_IMAGE_URL from  a sylva unit via kustomization.

@CrystalChun This base/ and overlays/ocp/ folder are meant for internal use only, right? maybe they can be re-organized for some e2e CI test and used to deploy all manifests required for the CAPI provider to emulate how an end user will deploy the manifests on a k8s or OCP cluster. But that's outside of the scope of this PR and we can discuss that later.

bootstrap/config/manager/base/deployment.yaml
```diff
@@ -73,6 +60,13 @@ spec:
         image: controller:latest
         imagePullPolicy: Always
         name: manager
+        env:
```
rccrdpccl, line 63: Wouldn't it be easier if we have the base layer without these envs, and just add them in the overlays? This way if we add other env vars the order would not matter.
Also, we should probably disable this behaviour by default and assume public addresses. If that's not the case the user should know and configure the provider accordingly (I know the main use case is with this feature enabled, but I feel we should enable it when we declare what address we have, as by default we would need to fail the controller if it's enabled but not internal namespace/service it's defined)

bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -18,23 +18,39 @@ package controller
...
 
+type InfraEnvControllerConfig struct {
+	// UseInternalImageURL when set to false means that we'll use the InfraEnv's iso download URL
+	// as is. When set to true, it'll use the assisted-image-service's internal IP as part of the
+	// download URL.
+	UseInternalImageURL bool `envconfig:"USE_INTERNAL_IMAGE_URL" default:"true"`
+	// ImageServiceName is the Service CR name for the assisted-image-service
+	ImageServiceName string `envconfig:"IMAGE_SERVICE_NAME" default:"assisted-image-service"`
+	// ImageServiceNamespace is the namespace that the Service CR for the assisted-image-service is in
+	ImageServiceNamespace string `envconfig:"IMAGE_SERVICE_NAMESPACE" default:"assisted-installer"`
```
rccrdpccl, line 46: Should this have a default from within the code? Normally we deploy this way, but can we really assume it from here? Wouldn't it be safer to force the use specify this?

bootstrap/config/manager/base/deployment.yaml
jianzzha: Regardless what default value to choose, the usage of the env var needs to be documented in the  README.md.

bootstrap/config/manager/overlays/ocp/kustomization.yaml
jianzzha: I think the intention of kustomize folder is for users who has cloned the source repo to be able to use kubectl apply -k to deploy the capi pods and resources in one step. The root level README.md does mention kubectl apply -k config/samples/. So I guess the purpose of the base/ and overlays/ folder here will be eventually used for that purpose? @CrystalChun @rccrdpccl

bootstrap/internal/controller/infraenv_controller.go
carbonin: This is part of the reason I asked @jianzzha to review.

I don't know what we will "normally" be doing.
When assisted is deployed through MCE it won't be in the `assisted-installer` namespace, but when we deploy it for dev, it is. I'm also not sure what namespace whatever MCE distribution we use for sylva will use. I don't know what to consider "normal" in this case.?

jianzzha: Sylva is referencing  https://github.com/openshift/assisted-service/tree/master/config/default kustomize folder at this moment, so  'assisted-installer' namespace is set by that kustomize. 

I'm assuming that MCE can kustomize the IMAGE_SERVICE_NAMESPACE if it needs to?

carbonin: > I'm assuming that MCE can kustomize the IMAGE_SERVICE_NAMESPACE if it needs to?

This is what I'm trying to ensure. MCE does not use kustomize when deploying assisted. I assume it also won't use kustomize when deploying this CAPI provider (if it is even going to be responsible for that). MCE pulls the files from the rendered operator bundle and applies those.

carbonin: But the open question (for me at least) is still if MCE is going to deploy the CAPI providers or if they're being installed through some other means.

rccrdpccl: When MCE deploys this we'll be under OCP, so public route can be used, thus turning off this hack
I think independently of what we would "normally" do, the safer option is to default to public route (advertised), and only when specifically overriding it it should be set. Of course we should reflect this in the docs as suggested.
We could also have a sample overlay already prepared where users could apply it with current default (basically moving defaults from code to a sample kustomize directory).

carbonin: >When MCE deploys this we'll be under OCP

No? Wasn't part of this effort to get MCE to install on vanilla kube as well?

bootstrap/config/manager/base/deployment.yaml
CrystalChun: I did initially try that (no env vars in the base and add them in overlay) but kustomize acted weird with it :/ I might've been patching it incorrectly though

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: Discussed offline with Nick that we should just look for assisted service in the same namespace as where the bootstrap controller is running since we have control over where this controller runs. 

In other words, we'll deploy this controller in the same namespace that assisted service runs.

jianzzha: This is fine. We have a kustomize sylva unit for both the infra operator and capi provider and can make sure they run in the same name space.

rccrdpccl: > In other words, we'll deploy this controller in the same namespace that assisted service runs.
I am not sure I agree with this. What is the advantage over the other possible solutions?

I can see the following scenarios:
* when override behaviour is disabled, we use public route: no need to know where assisted is runnins
* when override behaviour is enabled, we could do one of the following:
  * the user needs to specify where assisted is running
  * the user must deploy assisted and CAPI controllers in the same namespace

I think the current proposed default it's the least flexible scenario.


The advantages I see as leaving as default the public route is that:
* this should be the desired behaviour: reading from advertised address. OCP will work this way, and I believe if this is the way to go for the future, we should eventually advertise the private address. The current solution is a workaround.
 
We can always override and eventually stop overriding from Sylva unit.

Is there something that I am missing here?

carbonin: I suggested looking in the same namespace as a more sensible default than hardcoding `assisted-installer`.

I think we should still expose the env var as overrideable, but it was my understanding that we didn't have a good way to do that from kustomize without hardcoding other namespaces.

As for where these components are running we only have three cases:
1. Everything installed from MCE -> everything runs in the same namespace
2. Assisted installed from MCE, CAPI provider installed by user -> doc to run CAPI provider in assisted namespace
3. User installs everything -> doc to run everything in the same namespace.

From what I could tell from the discussion around kustomize overrides for the namespace this seems more reliable.

rccrdpccl: I see and agree that running everything in the same namespace might be more sensible than hardcoding assisted's namespace. What about the following:
* Default to USE_INTERNAL_IMAGE_URL to false: this will use public route, where all of this is a non-issue and it would be the desired behaviour
* Default IMAGE_SERVICE_NAMESPACE to empty: when empty it will mean same namespace (if override is turned on), however it can be overridden if need be.

In Sylva's "upstream" case, we'd only need to add one env var through kustomization (USE_INTERNAL_IMAGE_URL=true) and deploy in the same namespace.
In MCE case, everything would be deployed in the same namespace, but no overrides needed as we'd go through the advertised public address.

Basically my point is that the "workaround" (USE_INTERNAL_IMAGE_URL) should not be default behaviour. Would you agree?

carbonin: >In MCE case, everything would be deployed in the same namespace, but no overrides needed

You're still assuming whenever we run with MCE we're going to be in OCP and I'm not sure yet that this will always be true.
But that said, I beleive MCE's installer can set env vars when it deploys operators so we'd have to get them to set this in the OCP/kube case no matter what we choose so I don't think it's an issue.

Plus we're still a bit away from having MCE install this provider so we can probably deal with this when it's actually an issue.

rccrdpccl: Wouldn't it be easier to just remove any "workaround" kustomization rather than having to change the upstream manifest, when/if we'll finally release that?
I feel we are making a workaround the default behaviour making some assumptions (what namespace we run what component) on the way. Shouldn't we allow the user to have this workaround and make the assumptions/decisions themselves?

carbonin: Had a discussion offline, the decision was to remove the default value for the image service namespace. If provided we will use it to look up the service, if not we will look for the service in the namespace the provider is running in.

We will also default `USE_INTERNAL_IMAGE_URL` to `false` and the user will need to change that if they want/need to.

bootstrap/internal/controller/infraenv_controller.go
CrystalChun: Changed it back to default to false as per conversation https://github.com/openshift-assisted/cluster-api-agent/pull/51#discussion_r1704478929

bootstrap/internal/controller/infraenv_controller.go
```diff
@@ -18,23 +18,40 @@ package controller
...
 
+type InfraEnvControllerConfig struct {
+	// UseInternalImageURL when set to false means that we'll use the InfraEnv's iso download URL
+	// as is. When set to true, it'll use the assisted-image-service's internal IP as part of the
+	// download URL.
+	UseInternalImageURL bool `envconfig:"USE_INTERNAL_IMAGE_URL" default:"false"`
+	// ImageServiceName is the Service CR name for the assisted-image-service
+	ImageServiceName string `envconfig:"IMAGE_SERVICE_NAME" default:"assisted-image-service"`
+	// ImageServiceNamespace is the namespace that the Service CR for the assisted-image-service is in
+	ImageServiceNamespace string `envconfig:"IMAGE_SERVICE_NAMESPACE" default:""`
```
carbonin, line 47: Super minor but we probably don't need the `default` at all here, right?

CrystalChun: Oh yep good point! Removing it

carbonin:
Looks good to me. I'll let @rccrdpccl give it the final lgtm though

rccrdpccl:
Could we please add some documentation about how would we use this feature and why?

CrystalChun:
> Could we please add some documentation about how would we use this feature and why?

Sorry missed it originally, just added it!

docs/deploy.md
```diff
@@ -0,0 +1,66 @@
+# Deploying Cluster API Provider Openshift BareMetal
```
rccrdpccl, line 1: I think this should probably be a section of the main README.md

docs/deploy.md
```diff
@@ -0,0 +1,66 @@
...
+To deploy this provider on vanilla Kubernetes, the following must be done:
+
+### Prerequisites
+
+A vanilla Kubernetes such as [Kind](https://kind.sigs.k8s.io/) must be deployed and you must be authenticated to the cluster.
+
+The following services are required on your cluster before installing this provider.
+
+1. Install [Assisted-Service operator](https://github.com/openshift/assisted-service/blob/master/docs/dev/operator-on-kind.md)
+2. Install [CAPI](https://cluster-api.sigs.k8s.io/user/quick-start.html)
```
rccrdpccl, line 16: We already have some instructions in the readme about installing the providers, can we integrate those docs with these new instructions?

docs/deploy.md
```diff
@@ -0,0 +1,66 @@
...
+3. Install [CAPM3](https://book.metal3.io/capm3/installation_guide)
+
+The following CLI tools are required:
+
+1. [kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/)
```
rccrdpccl, line 21: Requirements are also stated in README.md

docs/deploy.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Deploy providers
+
+Clone this github repository
+
+```bash
+git clone http://github.com/openshift-assisted/cluster-api-agent
+```
+
+Append the following to the bottom of the manager [kustomization.yaml file](../config/manager/kustomization.yaml)
```
rccrdpccl, line 32: Should we suggest the remote resource approach instead of checking out the whole repo?
i.e.

```
cat <<EOF > mytest/myfile
heredoc> apiVersion: kustomize.config.k8s.io/v1beta1
heredoc> kind: Kustomization
heredoc> metadata:
  name: internal-image-url
resources:
- https://raw.githubusercontent.com/openshift-assisted/cluster-api-agent/master/bootstrap/config/manager/manager.yaml
...
```

CrystalChun: Good idea, modified!

docs/deploy.md
CrystalChun: Gotcha, yes tried to do so but since we're suggesting a kustomize with a patch, these instructions/prerequisites mostly stayed the same 😅 do you know how to do it with clusterctl? 


Also would you like assisted service as a prerequisite for the general deployment instructions? I was assuming we're targeting OCP so assisted service would already be deployed and need not be listed as a general prerequisite. Only listed if it's vanilla kube. Let me know your thoughts!

docs/deploy.md
CrystalChun: Moved it, thanks!

docs/deploy.md
CrystalChun: Gotcha, thanks!

docs/deploy.md
rccrdpccl: AFAIK we cannot do this with clusterctl, but I meant maybe we can have one section explaining how to install, with all the variants.

PR #52, opened 2024-07-22
adds code commentary and improves a log message

mhrivnak:


bootstrap/api/v1alpha1/agentbootstrapconfig_types.go
```diff
@@ -87,6 +87,7 @@ type AgentBootstrapConfigStatus struct {
 	// AgentRef references the agent this agent bootstrap config has booted
 	AgentRef *corev1.LocalObjectReference `json:"agentRef,omitempty"`
 
+	// TODO: should this be on the spec?
```
CrystalChun, line 90: I believe we initially moved it here to prevent the user from setting this field. 
What are the reasons to move it to spec?

bootstrap/internal/controller/agent_controller.go
```diff
@@ -147,6 +149,10 @@ func (r *AgentReconciler) findBMH(
 		return nil, errors.New("agent doesn't have inventory yet")
 	}
 
+	// TODO maybe we can find a better way than listing and scanning every BMH
+	// that exists, such as explicitly watching BMHs and setting up a
+	// FieldIndexer.
+	// Example: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
```
CrystalChun, line 155: Good idea! Will look into this, thank you!!

bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -432,12 +434,14 @@ func (r *AgentBootstrapConfigReconciler) setMetal3MachineImage(
 			return err
 		}
 		log.V(logutil.TraceLevel).
-			Info("Added ISO URLs to metal3 machines", "machine", metal3Machine.Name, "namespace", metal3Machine.Namespace)
+			Info("Added ISO URL to metal3 machine", "machine", metal3Machine.Name, "namespace", metal3Machine.Namespace)
```
CrystalChun, line 437: Good catch, thank you!!

bootstrap/api/v1alpha1/agentbootstrapconfig_types.go
mhrivnak: It's desired state. We want the URL to be used in the bootstrap process. And of course in a declarative API, we "spec"ify our desired state in the `spec`. :)

The status should be reserved for observations about the world that were taken the last time the owning controller reconciled this resource.

CrystalChun: Ah I see! That makes sense, thank you for the explanation! CC @rccrdpccl

bootstrap/internal/controller/agentbootstrapconfig_controller.go
```diff
@@ -553,6 +559,11 @@ func (r *AgentBootstrapConfigReconciler) SetupWithManager(mgr ctrl.Manager) erro
 		).
 		Watches(
 			&hivev1.ClusterDeployment{},
+			// TODO does this depend on the name and namespace for the
+			// ClusterDeployment being the same as the name and namespace of the
+			// ABC? Could we put an owner reference on the CD instead and use
```
rccrdpccl, line 564: Yes we definitely need to change a few naming convenction, ClusterImageSet is another example.

PR #53, opened 2024-08-06
fix rbac

rccrdpccl:


bootstrap/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/edge-infrastructure/openshift-capi-agent-bootstrap
-  newTag: latest
+  newName: quay.io/rh_ee_rpiccoli/openshift-capi-agent-bootstrap
```
carbonin, line 7: Probably shouldn't commit this.

rccrdpccl: :facepalm:

adriengentil (approved):

PR #76, opened 2025-01-31
NO-ISSUE: Update README.md and add CONTRIBUTING.md

CrystalChun:
Update README to include additional prerequisites for running the e2e tests and add file for contributing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,23 +217,34 @@ kubectl apply -f https://raw.githubusercontent.com/<org>/cluster-api-agent/<tag 
 
 E2E tests can be executed through ansible tasks in a target remote host.
 
-The following env vars are needed:
-`SSH_KEY_FILE` key file to access the remote host
-`REMOTE_HOST` remote host where to execute the tests
-`PULLSECRET` base64 encoded pullsecret to inject into the tests
+#### Prerequisites
+
+Export the following env vars:
+
+`SSH_KEY_FILE` path to the private SSH key file to access the remote host
+`SSH_AUTHORIZED_KEY` value of your public SSH key which is used to access the hosts used for deploying the workload cluster
+`REMOTE_HOST` remote host name where to execute the tests: `root@<REMOTE_HOST>`
+`PULLSECRET` base64-encoded pull secret to inject into the tests
+`DIST_DIR` is this repository directory `/dist` i.e. `$(pwd)/dist`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -217,23 +217,34 @@ kubectl apply -f https://raw.githubusercontent.com/<org>/cluster-api-agent/<tag
...
-`SSH_KEY_FILE` key file to access the remote host
-`REMOTE_HOST` remote host where to execute the tests
-`PULLSECRET` base64 encoded pullsecret to inject into the tests
+#### Prerequisites
 
-Then we can run:
+Export the following env vars:
+
+`SSH_KEY_FILE` path to the private SSH key file to access the remote host
+`SSH_AUTHORIZED_KEY` value of your public SSH key
```
rccrdpccl, line 225: Used to access cluster nodes, should we specify it in the readme?

README.md
```diff
@@ -217,23 +217,34 @@ kubectl apply -f https://raw.githubusercontent.com/<org>/cluster-api-agent/<tag
...
+`REMOTE_HOST` remote host name where to execute the tests: `root@<REMOTE_HOST>`
+`PULLSECRET` base64-encoded pull secret to inject into the tests
+`DIST_DIR` is this repository directory `/dist` i.e. `$(pwd)/dist`
+`CONTAINER_TAG` set this to "local"
```
rccrdpccl, line 229: Should we change the behaviour so that the user can decide what value this should have?

README.md
CrystalChun: Yeah good idea!

README.md
CrystalChun: Ideally, yes! I do think we should have this documented for now since it is currently what's needed to run the ansible playbook, right?

rccrdpccl: Sure, I meant since it's a terrible way of doing it, we could've made it better while documenting :)
but I'm ok just documenting as is and improve it in the future

The issue is that [right now it can only be set to local](https://github.com/openshift-assisted/cluster-api-agent/blob/master/test/ansible/run_test.yaml#L99)

CrystalChun: Oh, I see what you mean. I thought you were asking to make everything configurable and didn't realize you were talking about just `CONTAINER_TAG`. Sorry! 

Sure we can definitely make `CONTAINER_TAG` a user-defined value! Adding to this PR in a moment

PR #77, opened 2025-02-11
add distributionVersion field to controlplane

rccrdpccl:
Added spec.distributionVersion field, see https://github.com/kubernetes-sigs/cluster-api/issues/11816
Added status.version field. This should reflect K8s version, which we are extracting from the release image payload

This field changes the controlplane API, so we need to release a new version: v1alpha2


controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
```diff
@@ -31,13 +31,21 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+type mockOpenShiftVersioner struct {
+}
+
+func (m *mockOpenShiftVersioner) GetK8sVersionFromReleaseImage(ctx context.Context, releaseImage string, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) (*string, error) {
+	version := "1.30.0"
```
CrystalChun, line 45: Are you planning on making this a todo to get the version from the release image?

controlplane/api/v1alpha2/groupversion_info.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 // Package v1beta1 contains API Schema definitions for the controlplane v1beta1 API group
```
CrystalChun, line 17: Oops looks like we missed this comment the first time 
```suggestion
// Package v1alpha2 contains API Schema definitions for the controlplane v1alpha2 API group
```

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
rccrdpccl: This is the mock/dummy in the test, we are already  - I can add an assertion making sure status.version is 1.30.0
the real thing is happening here https://github.com/openshift-assisted/cluster-api-agent/blob/distribution-version/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go#L172

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -113,33 +114,33 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
...
+		controllerutil.AddFinalizer(oacp, acpFinalizer)
 	}
 
-	acpVersion, err := semver.NewVersion(acp.Spec.Version)
+	acpVersion, err := semver.NewVersion(oacp.Spec.DistributionVersion)
 	if err != nil {
-		log.Error(err, "invalid OpenShift version", "version", acp.Spec.Version)
-		return ctrl.Result{}, err
+		// we accept any format (i.e. latest)
+		log.V(logutil.WarningLevel).Info("invalid OpenShift version", "version", oacp.Spec.DistributionVersion)
```
CrystalChun, line 135: Will having an error here still be ok for the rest of the reconcile function or would it be better to stop reconcile here?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -455,3 +458,34 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensurePullSecret(
 	acp.Spec.Config.PullSecretRef = &corev1.LocalObjectReference{Name: secret.Name}
 	return nil
 }
+
+func NewOpenShiftVersion(client client.Client) *OpenShiftVersion {
```
CrystalChun, line 462: What do you think about having this and its functions in its own file?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
rccrdpccl: If we stop reconcile here, we wouldn't be able to set `latest` as version

rccrdpccl:
Tests and linter failing because of release golang version (will change before merging), passing in local environment

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
CrystalChun: Ah gotcha, ok

PR #78, opened 2025-02-12
MGMT-19876: Detect upgrade

CrystalChun:
Adds the status field `distributionVersion` which is set to the OCP version in the workload cluster's cluster version when the workload cluster is installed. 

Determines if an upgrade is requested by comparing the `distributionVersion` spec field to the `distributionVersion` status field.


controlplane/internal/workloadclient/workload_client.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil, errors.Wrapf(err, "failed to get clientconfig from kubeconfig data")
+	}
+
+	restConfig, err := clientConfig.ClientConfig()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get restconfig for kube client")
+	}
+
+	schemes := runtime.NewScheme()
+	utilruntime.Must(configv1.Install(schemes))
```
rccrdpccl, line 24: Should we return an error instead of panicking here?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -175,9 +175,39 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
...
 	}
 	oacp.Status.Version = k8sVersion
+	upgradeRequested := determineUpgrade(ctx, oacp)
+	if upgradeRequested {
+		//TODO: Handle upgrade request
+	}
 	return ctrl.Result{}, r.reconcileReplicas(ctx, oacp, cluster)
 }
 
+func determineUpgrade(ctx context.Context, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) bool {
```
rccrdpccl, line 185: should we rename the method to `isUpgradeRequested`? We could then do
```
if isUpgradeRequested(ctx, oacp) {
  // handle upgrade
}
```

controlplane/internal/controller/agentclusterinstall_controller.go
```diff
@@ -79,6 +82,16 @@ func (r *AgentClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 			log.Error(err, "failed setting OpenshiftAssistedControlPlane to ready")
 			return ctrl.Result{}, err
 		}
+		//TODO: determine a better condition to add this
+		kubeConfigSecret, err := r.getACIKubeconfig(ctx, aci, acp)
+		if err != nil {
+			log.Error(err, "failed to get Kubeconfig secret for workload client")
+			return ctrl.Result{}, err
+		}
+		if err := setControlPlaneVersion(ctx, kubeConfigSecret, &acp); err != nil {
```
rccrdpccl, line 91: If we just set the ready condition true of "WorkloadKubeconfig available" to the OACP, we could then move this logic to the controlplane controller and trigger it by the condition being ready. We could also remove the code retrieving the kubeconfig as it's already done in the reconcile function - we can retrieve it from the well known secret for kubeconfig (our secret, not the ACI's) - WDYT?

rccrdpccl (commented):
nice approach!

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
CrystalChun: Good idea!

controlplane/internal/controller/agentclusterinstall_controller.go
CrystalChun: Great idea! I'll implement that since retrieving the kubeconfig here is kind of clunky and the condition I have for setting that version is unideal too.

CrystalChun:
/close
Will open a new PR against master

controlplane/internal/workloadclient/workload_client.go
CrystalChun: Whoops! Good call

PR #80, opened 2025-02-13
MGMT-19876: Detect upgrade

CrystalChun:
Adds the status field `distributionVersion` which is set to the OCP version in the workload cluster's cluster version when the workload cluster is installed. 

Determines if an upgrade is requested by comparing the `distributionVersion` spec field to the `distributionVersion` status field.

Supersedes https://github.com/openshift-assisted/cluster-api-agent/pull/78

### DIFF
--- a/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
+++ b/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
@@ -182,6 +182,10 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
 	k8sVersion, err := r.OpenShiftVersion.GetK8sVersionFromReleaseImage(ctx, releaseImage, oacp)
 	markKubernetesVersionCondition(oacp, err)
 	oacp.Status.Version = k8sVersion
+
+	if isUpgradeRequested(ctx, oacp) {
+		//TODO: Handle upgrade request
+	}
 	return ctrl.Result{}, r.reconcileReplicas(ctx, oacp, cluster)
 }
 
@@ -198,6 +202,32 @@ func markKubernetesVersionCondition(oacp *controlplanev1alpha2.OpenshiftAssisted
 	} else {
 		conditions.MarkTrue(oacp, controlplanev1alpha2.KubernetesVersionAvailableCondition)
 	}
+}
+
+func isUpgradeRequested(ctx context.Context, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) bool {
+	log := ctrl.LoggerFrom(ctx)
+	oacpDistVersion, err := semver.NewVersion(oacp.Spec.DistributionVersion)
+	if err != nil {
+		log.Error(err, "failed to detect OpenShift version from ACP spec", "version", oacp.Spec.DistributionVersion)
+		return false
+	}
+
+	upgrade := false
+	if oacp.Status.DistributionVersion == "" {
+		return false
+	}
+	currentOACPDistVersion, err := semver.NewVersion(oacp.Status.DistributionVersion)
+	if err != nil {
+		log.Error(err, "failed to detect OpenShift version from ACP status", "version", oacp.Spec.DistributionVersion)
+		return false
+	}
+
+	if oacpDistVersion.Compare(*currentOACPDistVersion) > 0 {
+		log.Info("Upgrade detected, new requested version is greater than current version",
+			"new requested version", oacpDistVersion.String(), "current version", currentOACPDistVersion.String())
+		upgrade = true
+	}
+	return upgrade
 }
 
 // Ensures dependencies are deleted before allowing the OpenshiftAssistedControlPlane to be deleted
--- a/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
+++ b/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
@@ -186,11 +186,11 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
 	k8sVersion, err := r.OpenShiftVersion.GetK8sVersionFromReleaseImage(ctx, releaseImage, oacp)
 	markKubernetesVersionCondition(oacp, err)
 	oacp.Status.Version = k8sVersion
-	if oacp.Status.DistributionVersion, err = r.getWorkloadClusterVersion(ctx, oacp); err != nil {
+	if oacp.Status.DistributionVersion, err = GetWorkloadClusterVersion(ctx, r.Client, r.WorkloadClusterClientGenerator, oacp); err != nil {
 		log.Error(err, "failed to set the openshift version in the control plane status")
 	}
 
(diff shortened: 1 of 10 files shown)


controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -182,6 +186,12 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
 	k8sVersion, err := r.OpenShiftVersion.GetK8sVersionFromReleaseImage(ctx, releaseImage, oacp)
 	markKubernetesVersionCondition(oacp, err)
 	oacp.Status.Version = k8sVersion
+	if err := r.setOpenshiftVersionInStatus(ctx, oacp); err != nil {
```
rccrdpccl, line 189: I think `setOpenshiftVersionInStatus` doesn't clearly say where this version is coming from. What about:
```
oacp.Status.DistributionVersion, err = r.getWorkloadClusterVersion(...)
```

Wouldn't this be clearer?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -492,3 +528,69 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensurePullSecret(
...
+
+func (r *OpenshiftAssistedControlPlaneReconciler) getKubeconfigSecret(ctx context.Context, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) *corev1.Secret {
+	// Retrieve clusterdeployment
+	cd := &hivev1.ClusterDeployment{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: oacp.Status.ClusterDeploymentRef.Namespace, Name: oacp.Status.ClusterDeploymentRef.Name}, cd); err != nil {
+		return nil
+	}
+	if cd.Spec.ClusterInstallRef == nil {
+		return nil
+	}
```
rccrdpccl, line 583: We already handle this: https://github.com/openshift-assisted/cluster-api-agent/blob/master/controlplane/internal/controller/agentclusterinstall_controller.go#L138-L150

I think we can extract common code, and we would already know what is the expected path for the kubeconfig (the one we copy from ACI's). It will save us 2 GET and simplify the code

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
CrystalChun: Yeah good idea!

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
CrystalChun: Great idea! Extracted it out to a util function. What do you think?

CrystalChun:
> This looks great 😍 could you please add some tests for it? 🙏

Thanks!! Added a couple for the main functionality this PR adds, but can definitely add more in a later PR?

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
```diff
@@ -273,6 +274,123 @@ var _ = Describe("OpenshiftAssistedControlPlane Controller", func() {
...
+				conditions.MarkFalse(
+					openshiftAssistedControlPlane,
+					controlplanev1alpha2.KubeconfigAvailableCondition,
+					controlplanev1alpha2.KubeconfigUnavailableFailedReason,
+					clusterv1.ConditionSeverityInfo,
+					"error retrieving Kubeconfig %v", fmt.Errorf("secret does not exist"),
+				)
+
+				By("confirming an error is returned when calling getWorkloadClusterVersion")
+				_, err := controllerReconciler.getWorkloadClusterVersion(ctx, openshiftAssistedControlPlane)
```
rccrdpccl, line 336: Ideally we should not test private methods. However, this doesn't necessarily belong to the controller package, so we could extract it into its own and leave the test virtually unchanged.

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
```diff
@@ -273,6 +274,123 @@ var _ = Describe("OpenshiftAssistedControlPlane Controller", func() {
...
+				openshiftAssistedControlPlane := getOpenshiftAssistedControlPlane()
+				openshiftAssistedControlPlane.Status.DistributionVersion = "4.18.0"
+				Expect(isUpgradeRequested(context.Background(), openshiftAssistedControlPlane)).To(BeFalse())
+			})
+		})
+		When("openshiftassistedcontrolplane's requested upgrade is greater than the current workload cluster's version", func() {
+			It("returns true", func() {
+				openshiftAssistedControlPlane := getOpenshiftAssistedControlPlane()
+				openshiftAssistedControlPlane.Status.DistributionVersion = "4.11.0"
+				Expect(isUpgradeRequested(context.Background(), openshiftAssistedControlPlane)).To(BeTrue())
```
rccrdpccl, line 388: Same here: we are testing private methods.

How would the upgrade look like? If there is an updater or something, we could mock and assert it has been called. In case this approach won't work, we can leave the test for now and substitute it when we'll integrate the actual logic of upgrades (and make some assertion on it). WDYT?

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
CrystalChun: Extracted it! How does it look?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -200,6 +212,32 @@ func markKubernetesVersionCondition(oacp *controlplanev1alpha2.OpenshiftAssisted
 	}
 }
 
+func IsUpgradeRequested(ctx context.Context, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) bool {
```
rccrdpccl, line 215: Since we've made it public, I think we should create a package (or find a suitable one) for this and keep the controller package slimmer -same goes for GetWorkloadClusterVersion. WDYT?

CrystalChun: Good idea! Moved it out!

controlplane/internal/upgrade/upgrade.go
```diff
@@ -0,0 +1,105 @@
...
+		return "", err
+	}
+
+	var clusterVersion configv1.ClusterVersion
+	if err := workloadClient.Get(ctx, types.NamespacedName{Name: "version"}, &clusterVersion); err != nil {
+		err = errors.Join(err, fmt.Errorf(("failed to get ClusterVersion from workload cluster")))
+		return "", err
+	}
+
+	return clusterVersion.Status.Desired.Version, nil
```
rccrdpccl, line 63: This is probably `.status.history[?(@.state=="Completed")].version` https://github.com/openshift/api/blob/744790f2cff777b1bb29bdccce10e4e84cff0a69/config/v1/types_cluster_version.go#L22

PR #82, opened 2025-02-25
MGMT-19876: Set version based on cluster version history

CrystalChun:
If history in ClusterVersion is available, set the control plane version based off of the completed version. Also logs when a requested upgrade is not actually an upgraded version.

/cc @rccrdpccl


rccrdpccl:
Sorry @CrystalChun I've somehow missed this PR. It seems most of the logic has already been merged, do you think there's something that could yet be added?

PR #83, opened 2025-02-28
handle cluster upgrades with feedback from CVO

rccrdpccl:
Add dependency injection to make unit testing easier on:
* upgrade logic
* container image manipulation

Add logic for upgrade feedback
Add e2e tests for upgrades


controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -181,22 +185,119 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
...
+			RequeueAfter: 1 * time.Minute,
+		}, nil
+	}
+
+	if isWorkloadClusterRunningDesiredVersion(oacp) && !isUpdateInProgress {
+		log.V(logutil.WarningLevel).Info("Cluster is now running expected version, upgraded completed")
+
+		conditions.MarkTrue(oacp, controlplanev1alpha2.UpgradeCompletedCondition)
+		return ctrl.Result{}, nil
+	}
```
CrystalChun, line 268: The `isWorkloadClusterRunningDesiredVersion` function https://github.com/openshift-assisted/cluster-api-agent/pull/83/files#diff-12f36ba5c9185bd1b2b7d0fc80636925dbd3f20b2509055100d988ebf99e0810R298

Doesn't tell us whether the requested version from `spec.distributionVersion` is an upgrade or downgrade compared to what the workload cluster is running. Should we be taking that into account?

controlplane/internal/upgrade/upgrade.go
```diff
@@ -2,142 +2,202 @@ package upgrade
...
+// Returns true if the desired version matches the cluster's desired version, false otherwise. If any error occurs while
+// performing this operation, it will be returned
+func (u *OpenshiftUpgrader) IsDesiredVersionUpdated(ctx context.Context, desiredVersion string) (bool, error) {
+	clusterVersion, err := u.getClusterVersion(ctx)
 	if err != nil {
-		err = errors.Join(err, fmt.Errorf("failed to extract kubeconfig from secret %s", kubeconfigSecret.Name))
-		return nil, err
+		return false, err
 	}
+	return isUpdateInProgress(clusterVersion), nil
```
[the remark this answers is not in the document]
rccrdpccl: :facepalm: https://github.com/openshift-assisted/cluster-api-agent/pull/86

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
rccrdpccl: IIRC OpenShift does not support downgrade, we can forward this and let CVO give us feedback about it.
What we can do is to try this scenario and make sure we do provide feedback. WDYT?

PR #84, opened 2025-03-05
MGMT-19400 - create tag management flow to tag assisted installer components according to capbcoa versions

omer-vishlitzky:
https://issues.redhat.com/browse/MGMT-19400

## Summary by CodeRabbit

- **New Features**
	- Launched enhancements for automated release candidate management, including streamlined workflows for testing, tagging, and version discovery.
	- Enabled dynamic configuration using environment variables for versioning and infrastructure deployments.
	- Introduced a new test runner for Ansible tests related to release candidates.
	- Added functionality for managing Git tags in repositories.
	- Implemented a component scanner for version information and associated Docker images.

- **Tests**
	- Added comprehensive test suites verifying the new automation, tagging, and version discovery functionalities.
	- Included unit tests for the Ansible test runner and tag reconciler.

- **Chores**
	- Updated settings to ignore temporary cache directories.
	- Included new sample configuration files to support testing and version management.

### DIFF
--- a/test/playbooks/roles/system_dependencies/tasks/main.yaml
+++ b/test/playbooks/roles/system_dependencies/tasks/main.yaml
@@ -26,3 +26,4 @@
   ansible.builtin.pip:
     name:
       - kubernetes
+      - jmespath


scripts/tag_reconciler.py
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        try:
+            response = requests.get(url, headers=headers)
+            return response.status_code == 200
+        except requests.RequestException as e:
+            logger.error(f"Request error checking tag {tag} for {repo}: {e}")
+            return False
+
+    def create_tag(self, repo: str, commit_sha: str, tag: str) -> bool:
```
rccrdpccl, line 56: Why using raw http requests instead of PyGitHub? It will also be easier for authentication with a github APP

scripts/tag_reconciler.py
```diff
@@ -0,0 +1,173 @@
...
+
+        repositories = version.get("repositories", {})
+        if not repositories:
+            logger.warning(f"No repositories found for version {version_name}")
+            return True
+
+        for repo_name, repo_info in repositories.items():
+            commit_sha = repo_info.get("commit_sha", "")
+
+            if not re.match(r"^openshift/", repo_name) or not commit_sha:
```
rccrdpccl, line 125: Not critical as I am not foreseeing using a non openshift org component anytime soon, however why did we implement this limitation?

scripts/version_discovery.py
```diff
@@ -0,0 +1,292 @@
...
+                    "version": tag.replace("latest-", ""),
+                    "image_url": f"{image}:{tag}",
+                }
+            else:
+                logger.warning(f"No commit-based images found for {component_name}")
+
+    return repo_results
+
+
+def save_to_release_candidates(
```
rccrdpccl, line 238: Wouldn't a ReleaseCandidates class or something similar help? We split how to read and write in different files, which might be awkward to find and change.

scripts/version_discovery.py
```diff
@@ -0,0 +1,292 @@
...
+def process_repository(repo: str, config: RepositoryConfig) -> dict[str, ComponentInfo]:
+    repo_results: dict[str, ComponentInfo] = {}
+
+    tag_format = config.get("tag_format", "commit")
+    version_prefix = config.get("version_prefix", "v")
+    images = config.get("images", [])
+
+    # handle capi and capm3 tag format
+    if tag_format == "clusterctl":
+        logger.info(f"Checking {repo} using clusterctl-compatible versioning")
```
rccrdpccl, line 196: What does `clusterctl-compatible versioning` mean? What's the difference with regular versioning, and why `clusterctl` in the name?

scripts/test_runner.py
```diff
@@ -0,0 +1,399 @@
...
+        logger.info("Starting parallel test runner")
+        pending_indices, pending_snapshots = get_pending_snapshots()
+        if not pending_snapshots:
+            logger.info("No pending snapshots to process")
+            return
+
+        logger.info(f"Found {len(pending_snapshots)} pending snapshots to process")
+        results = []
+
+        with ThreadPoolExecutor(max_workers=len(pending_snapshots)) as executor:
```
rccrdpccl, line 365: Why parallel testing? Although it will definitely be faster, would the target machine have enough resources to run them altogether? Am I misunderstanding something here?

scripts/test_runner.py
```diff
@@ -0,0 +1,399 @@
...
+                return
+
+            with open(run_test_path, "w") as file:
+                local_yaml.dump(data, file)
+
+        except Exception as e:
+            logger.error(f"Error updating Ansible var {var_name}: {e}")
+            raise
+
+    def update_infrastructure_operator(self, component: str, image_url: str) -> None:
```
rccrdpccl, line 139: Wouldn't encapsulating this into the test be better? We can accept env vars with reasonable defaults in ansible  (in case of assisted components would default to not changing maybe?), and during the script we can setup as we are doing here (possibly just moving place for this piece of code with minor changes would do).
The advantage of this approach is that we could easily change versions of components just by exporting env vars:
```
ASSISTED_SERVICE_VERSION=0afd5315ee ansible-playbook ...
```
while right now this would be a feature unique to this tagging system.

rccrdpccl (changes requested):
Can you please add some basic testing for this functionality?

scripts/tag_reconciler.py
omer-vishlitzky: It was pretty straightforward so I didn't bother with trying pygithub actually. the operations here are relatively simple so i thought the dependency is unnecessary.
If you think we should use it for maintainability or something we can discuss this of course

scripts/tag_reconciler.py
omer-vishlitzky: the versions yaml file also contains the capi and capm3 repositories, and we don't want to tag those of course, so I added this check to ignore them

scripts/test_runner.py
omer-vishlitzky: You are not misunderstanding, and each test would be serial, but if we have multiple snapshots we want to test at once I thought that doing that in parallel would make things faster (as for resources, you said your machine has 400gb ram, I assume thats enough to run several installations in parallel but I might be wrong here)

scripts/version_discovery.py
omer-vishlitzky: yes that log is phrased horribly.
What I meant is that when searching for commits of assisted components we search for commit shas, but the capi and capm3
(which i called clusterctl because they are used by clusterctl in this line 
```
shell: "clusterctl init --core cluster-api:{{ capi_version }} --bootstrap - --control-plane - --infrastructure metal3:{{ capm3_version }}"
```
components use regular versioning (like v1.2.3).
it means we just need to distinguish exactly what versions we are looking for, commits (in case of assisted) or releases (in case of capi/capm3)

scripts/version_discovery.py
omer-vishlitzky: Yes you are right, I'll change that, thanks!

test/ansible/run_test.yaml
```diff
@@ -10,15 +10,26 @@
...
     src_dir: /tmp/capbcoa
     kind_cluster_name: capi-baremetal-provider
     ssh_authorized_key: "{{ lookup('ansible.builtin.env', 'SSH_AUTHORIZED_KEY') }}"
     pullsecret: "{{ lookup('ansible.builtin.env', 'PULLSECRET') }}"
     container_tag: "{{ lookup('ansible.builtin.env', 'CONTAINER_TAG', default='local')}}"
   tasks:
+  - name: prepare infrastructure-operator
+    delegate_to: localhost
+    shell: |
+      envsubst < "{{ playbook_dir }}/../e2e/manifests/infrastructure-operator/kustomization.yaml" > kustomization.yaml.new
```
rccrdpccl, line 25: nit: we migth want to save this as `/e2e/manifests/infrastructure-operator/kustomization.yaml.tpl`, so it's clear that it's not an usable kustomization file, and we would need only one step here.

Probably we could do the same with other files, but it'd be outside the scope of this PR.

rccrdpccl: Also, we might want to do this in the target machine instead, so we can leave local copy clean

scripts/tag_reconciler.py
rccrdpccl: I think the code would be much neater with PyGitHub, and also try to look up github app auth system, I believe it wouldn't be a simple static bearer token, which will complicate the code.

scripts/version_discovery.py
rccrdpccl: I think we can simplify and automatically detect if it's an hash or a tag (or maybe try to fetch hash, if error try tag?) this way it'd work for both  (i.e. I want to try an already tagged assisted-components and I don't want to use the sha or I want to try a specific sha in an upstream component)

scripts/test_runner.py
omer-vishlitzky: done

scripts/test_runner.py
```diff
@@ -0,0 +1,152 @@
...
+            return i, snapshot
+    return None, None
+
+
+def export_component_variables(snapshot: dict[str, Any]) -> None:
+    env_vars = {}
+    component_env_map = {
+        "kubernetes-sigs/cluster-api": "CAPI_VERSION",
+        "metal3-io/cluster-api-provider-metal3": "CAPM3_VERSION",
+        "openshift/assisted-service/assisted-service": "ASSISTED_SERVICE_IMAGE",
```
rccrdpccl, line 44: We probably should change also the operator image (kustomize images[0].digest), sorry if I didn't mention it earlier

scripts/tag_reconciler.py
```diff
@@ -0,0 +1,144 @@
...
+                logger.info(f"Tag {version_name} already exists for {base_repo}")
+                continue
+            if not self.create_tag(base_repo, commit_sha, version_name):
+                logger.error(f"Failed to create tag {version_name} for {base_repo}")
+                success = False
+
+        return success
+
+    def reconcile_tags(self) -> bool:
+        versions_data = self.read_versions_file()
```
rccrdpccl, line 111: Can we document somwehere the expected format of this file? With a test would be ideal, if complicated a README would do

scripts/test_runner.py
```diff
@@ -0,0 +1,152 @@
...
+    except Exception as e:
+        logger.error(f"Error updating snapshot status: {e}")
+
+
+def main() -> None:
+    try:
+        logger.info("Starting test runner")
+        data = read_release_candidates()
+
+        snapshot_index, snapshot = find_pending_snapshot(data)
```
rccrdpccl, line 129: I think we can do better than index here, maybe some name (date-based? ) that would make it easier to understand what's being processed.
Also please clarify what is the expected format (test or otherwise)

scripts/tag_reconciler.py
rccrdpccl: Actually I think defining a type for each version and the version file could be clearer.

As generic remark I think we need in the version (and candidate version) file the git repo/git ref pairs. This could come both as a tag or as a commit ref, so it'd be the same for owned and not owned repos.

Then we could use another utility to transform this values in images if need be (to fetch the right sha)

scripts/test_runner.py
omer-vishlitzky:
```
images:
  - name: quay.io/edge-infrastructure/assisted-service
    newName: quay.io/edge-infrastructure/assisted-service
    digest: ${ASSISTED_SERVICE_IMAGE} 
```

done

.gitignore
```diff
@@ -25,3 +25,4 @@ go.work
 *.swp
 *.swo
 *~
+scripts/__pycache__/*
```
rccrdpccl, line 28: Let's move all the scripts inside the `hack/` folder, as it's a common pattern in golang projects it would be a less surprising place to find them

scripts/tag_reconciler.py
```diff
@@ -0,0 +1,145 @@
...
+                logger.warning(f"Skipping repository {repo_name}")
+                continue
+
+            base_repo = self.extract_base_repo(repo_name)
+            if self.tag_exists(base_repo, version_name):
+                logger.info(f"Tag {version_name} already exists for {base_repo}")
+                continue
+            if not self.create_tag(base_repo, ref, version_name):
+                logger.error(f"Failed to create tag {version_name} for {base_repo}")
+                success = False
```
rccrdpccl, line 107: Can be extracted in another method for better readability

scripts/test_ansible_test_runner.py
```diff
@@ -0,0 +1,120 @@
...
+          generated_at: '2025-03-10T10:32:04.642635'
+          status: pending
+        components:
+          - repository: https://github.com/kubernetes-sigs/cluster-api
+            ref: v1.9.5
+            image_url:
+          - repository: https://github.com/metal3-io/cluster-api-provider-metal3
+            ref: v1.9.3
+            image_url:
+          - repository: assisted-service/assisted-service
```
rccrdpccl, line 94: Repository should be full nam,e of the repo, including github

scripts/test_ansible_test_runner.py
```diff
@@ -0,0 +1,120 @@
...
+            image_url: quay.io/edge-infrastructure/assisted-installer-agent:latest-cfe93a9779dea6ad2a628280b40071d23f3cb429
+          - repository: assisted-installer/assisted-installer-controller
+            ref: c389a38405383961d26191799161c86127451635
+            image_url: quay.io/edge-infrastructure/assisted-installer-controller:latest-c389a38405383961d26191799161c86127451635
+          - repository: openshift/assisted-installer/assisted-installer
+            ref: c389a38405383961d26191799161c86127451635
+            image_url: quay.io/edge-infrastructure/assisted-installer:latest-c389a38405383961d26191799161c86127451635
+    """
+    
+    with patch("builtins.open", mock_open(read_data=yaml_content)):
```
rccrdpccl, line 111: Wouldn't it better to actually have a test yaml file in some assets folder, so it would be more readable *and* we can actually read a file? This means we would need to parametrize the filename but I do not think that's an issue.

scripts/test_tag_reconciler.py
```diff
@@ -0,0 +1,104 @@
...
+          - repository: openshift/assisted-installer-agent/assisted-installer-agent
+            ref: cfe93a9779dea6ad2a628280b40071d23f3cb429
+            image_url: quay.io/edge-infrastructure/assisted-installer-agent:latest-cfe93a9779dea6ad2a628280b40071d23f3cb429
+          - repository: openshift/assisted-installer/assisted-installer-controller
+            ref: c389a38405383961d26191799161c86127451635
+            image_url: quay.io/edge-infrastructure/assisted-installer-controller:latest-c389a38405383961d26191799161c86127451635
+          - repository: openshift/assisted-installer/assisted-installer
+            ref: c389a38405383961d26191799161c86127451635
+            image_url: quay.io/edge-infrastructure/assisted-installer:latest-c389a38405383961d26191799161c86127451635
+    """
```
rccrdpccl, line 94: Same comments as candidate versions apply

scripts/version_discovery.py
```diff
@@ -0,0 +1,279 @@
...
+                        "repository": repo,
+                        "ref": commit_sha,
+                        "image_url": f"{image}:{tag}",
+                    })
+                else:
+                    logger.warning(f"No images found for {component_name}")
+        else:
+            logger.warning(f"No releases or tags found for {repo} and no images configured")
+
+        return repo_results
```
rccrdpccl, line 214: Can we simplify all those if-else and reduce nesting by returning early?

test/ansible/run_test.yaml
```diff
@@ -146,6 +146,15 @@
   - name: Debug tests output
     debug: "msg={{ loadbalancerip.stdout }}"
 
+  - name: prepare infrastructure-operator on target
+    shell: |
+      envsubst < "{{ src_dir }}/test/e2e/manifests/infrastructure-operator/kustomization.yaml.tpl" > "{{ remote_manifests_path }}/infrastructure-operator/kustomization.yaml"
```
rccrdpccl, line 151: Why output in `{{ remote_manifests_path }}`? BTW I think this might be unused, we can remove it

scripts/version_discovery.py
```diff
@@ -0,0 +1,279 @@
...
+
+    @staticmethod
+    def save_to_release_candidates(
```
rccrdpccl, line 217: I think we need to extract release candidate loading and saving.
I would recommend repository pattern, where we would have ReleaseCandidateRepository and a ReleaseCandidate object (structured or unstructured)
For example, something like:
```
rc_repo = ReleaseCandidateRepository("release_candidates.yaml)
rc_list = rc_repo.list() 
rc_list.append(ReleaseCandidate({}))
rc_repo.save_list(rc_list)
```

The advantage is that we centralize the structure of this file: right now it's dispersed in multiple files and it makes it more difficult to understa.d

scripts/test_ansible_test_runner.py
```diff
@@ -0,0 +1,120 @@
...
+            check=False, 
+            capture_output=False, 
+            text=True
+        )
+
+def test_read_release_candidates_success():
+    yaml_content = """
+    snapshots:
+      - metadata:
+          generated_at: '2025-03-10T10:32:04.642635'
```
rccrdpccl, line 85: we can leave generated_at, but I'd add a unique name/id (could be based on time, but since we have this field could be totally random now).

scripts/ansible_test_runner.py
```diff
@@ -0,0 +1,147 @@
...
+    except Exception as e:
+        logger.error(f"Error updating snapshot status: {e}")
+
+
+def main() -> None:
+    try:
+        logger.info("Starting test runner")
+        data: ReleaseCandidates = read_release_candidates()
+
+        timestamp, snapshot = find_pending_snapshot(data)
```
rccrdpccl, line 124: Again repo pattern would help here.
```
rc_list = rc_repo.find_release_candidates(status=pending)
# do my stuff with the pending candidates, modify the record(s)
rc_repo.save(id, rc_content)
```

rccrdpccl (changes requested):
Overall I think the architecture/flow looks great.

I've added some mainly aesthetic/readability comments

hack/test/assets/test_release_candidates.yaml
```diff
@@ -0,0 +1,27 @@
...
+        image_url: quay.io/edge-infrastructure/assisted-image-service:latest-2249c85d05600191b24e93dd92e733d49a1180ec
+      - repository: https://github.com/openshift/assisted-installer-agent
+        ref: cfe93a9779dea6ad2a628280b40071d23f3cb429
+        image_url: quay.io/edge-infrastructure/assisted-installer-agent:latest-cfe93a9779dea6ad2a628280b40071d23f3cb429
+      - repository: https://github.com/openshift/assisted-installer
+        ref: c389a38405383961d26191799161c86127451635
+        image_url: quay.io/edge-infrastructure/assisted-installer-controller:latest-c389a38405383961d26191799161c86127451635
+      - repository: https://github.com/openshift/assisted-installer
+        ref: c389a38405383961d26191799161c86127451635
+        image_url: quay.io/edge-infrastructure/assisted-installer:latest-c389a38405383961d26191799161c86127451635
```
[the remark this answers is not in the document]
rccrdpccl: This guy has a point, should we add `name` property for components? This will solve this issue, possibly the non-writable components one too

hack/shared_types.py
```diff
@@ -0,0 +1,32 @@
...
+    components: list[RepositoryInfo]
+
+class ReleaseCandidates(TypedDict):
+    snapshots: list[SnapshotInfo]
+
+class VersionEntry(TypedDict):
+    name: str
+    components: list[RepositoryInfo]
+
+class VersionsFile(TypedDict):
```
rccrdpccl, line 27: What's the difference between VersionFile and VersionEntry? The way I see it is that the file is just where we save a list of released versions. I think we should not have the word `file` in the name.
I think we might rethink how we represent those. What's the relation between the handled objects? The way I see it is:
* we discover a new "version candidate" or "snapshot"
* we validate such snapshot, if valid can be promoted to versions
* once added to versions, it will be reconciled

Can we find a way to represent this?

hack/test/assets/test_release_candidates.yaml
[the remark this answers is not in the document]
rccrdpccl: @omer-vishlitzky Could you please address this?

hack/version_discovery.py
```diff
@@ -0,0 +1,250 @@
...
+def main() -> None:
+    try:
+        logger.info("Starting component scan")
+        scanner = ComponentScanner()
+        results = scanner.scan_components()
+        if not results:
+            logger.error("No components found")
+            sys.exit(1)
+
+        scanner.save_to_release_candidates(results)
```
rccrdpccl, line 241: Let's try to follow the single responsibility principle: ComponentScanner should only scan. Then we can save its finding through another class, for example ReleaseCandidateRepository or some other abstraction on top of it.

I think at the moment the component scanner has too much knowledge about how we deal with release candidates (for example IMO it should not know the filename where we want to save to)

hack/test/assets/test_release_candidates.yaml
```diff
@@ -0,0 +1,27 @@
...
+        image_url: quay.io/edge-infrastructure/assisted-service:latest-76d29d2a7f0899dcede9700fc88fcbad37b6ccca
+      - repository: https://github.com/openshift/assisted-image-service
+        ref: 2249c85d05600191b24e93dd92e733d49a1180ec
+        image_url: quay.io/edge-infrastructure/assisted-image-service:latest-2249c85d05600191b24e93dd92e733d49a1180ec
+      - repository: https://github.com/openshift/assisted-installer-agent
+        ref: cfe93a9779dea6ad2a628280b40071d23f3cb429
+        image_url: quay.io/edge-infrastructure/assisted-installer-agent:latest-cfe93a9779dea6ad2a628280b40071d23f3cb429
+      - repository: https://github.com/openshift/assisted-installer
+        ref: c389a38405383961d26191799161c86127451635
+        image_url: quay.io/edge-infrastructure/assisted-installer-controller:latest-c389a38405383961d26191799161c86127451635
```
rccrdpccl, line 24: Thinking about image, I think we should definitely have the image repository (i.e. `quay.io/edge-infrastructure/assisted-installer-controller`), so that we can generalize how to search the latest image.
We might want to save the full image with digest/tag in "tested_image" or something like that

hack/release_candidates_repository.py
```diff
@@ -0,0 +1,163 @@
...
+            for d in self.components
+        }
+        set2 = {
+            (d.get("repository", ""), d.get("ref", ""), d.get("image_url", ""))
+            for d in other.components
+        }
+        return set1 == set2
+
+
+class ReleaseCandidatesRepository:
```
rccrdpccl, line 85: Repository pattern should only save and load to and from storage.
We might add options to retrieve filtered results, but we might want to push it to ReleaseCandidateCollection methods.

For the simplest repoistory, I envision something like this:

```
rcCollection = repo.list()
rc = rcCollection.find_by_status(status="pending")
new_status = run_test(rc)
rc.SetStatus(new_status)
rcCollection.upsert(rc)
# in this case, save deals with a list
repo.save(rcCollection)
```

for a more complex repository, we might want something like this:
```
opts = [FilterOption(field="status",value="pending")]

rc = repo.list(opts)[0] // check index/errors etc
new_status = run_test(rc)
rc.SetStatus(new_status)
# in this case, save deals with an item. Optionally we could take 2 params but we can infer ID by the object itself
repo.save(rc)
```

hack/shared_types.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+from typing import TypedDict
+
+class RepositoryInfo(TypedDict):
```
rccrdpccl, line 5: Should this be `Component`?

hack/version_discovery.py
```diff
@@ -0,0 +1,250 @@
...
+                    logger.info(f"Found matching image for {repo} at commit {sha[:8]}")
+                    return sha, tag
+
+            logger.warning(f"No commits found with matching images for {repo}/{image}")
+            return None
+        except Exception as e:
+            logger.error(f"Error finding commit with image for {repo}: {e}")
+            raise
+
+    def scan_components(self) -> list[RepositoryInfo]:
```
rccrdpccl, line 151: Should this return `ComponentCollection`

hack/version_discovery.py
```diff
@@ -0,0 +1,250 @@
...
+        if repo.save(candidate):
+            logger.info(
+                f"Saved new release candidate {candidate.metadata.get('id')} to {filename}"
+            )
+        else:
+            logger.info("No changes detected or error occurred, skipping update")
+
+
+def main() -> None:
+    try:
```
rccrdpccl, line 233: ReleaseCandidatesRepository could be initialized by factory method (so we'll load filename from env)

PR #90, opened 2025-03-06
MGMT-19878: Update control plane status for upgrade

CrystalChun:
Modifies status.ready to false when upgrade starts and changes it back to true when upgrade completes. Modifies updated replicas to the number of control plane nodes that are ready after upgrade completes.
These statuses are checked by Sylva when running an upgrade.

### DIFF
--- a/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
+++ b/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
@@ -234,6 +234,7 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
 	}
 	if isUpdateInProgress {
 		log.V(logutil.DebugLevel).Info("update is in progress...")
+		oacp.Status.Ready = false
 		conditions.MarkFalse(
 			oacp,
 			controlplanev1alpha2.UpgradeCompletedCondition,
@@ -246,6 +247,7 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
 	oacp.Status.DistributionVersion, err = upgrader.GetCurrentVersion(ctx)
 	if err != nil {
 		log.V(logutil.WarningLevel).Info("failed to get OpenShift version from ClusterVersion", "error", err.Error())
+		return ctrl.Result{}, err
 	}
 
 	// TODO: check for upgrade errors, mark relevant conditions
@@ -263,8 +265,12 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
 
 	if isWorkloadClusterRunningDesiredVersion(oacp) && !isUpdateInProgress {
 		log.V(logutil.WarningLevel).Info("Cluster is now running expected version, upgraded completed")
-
 		conditions.MarkTrue(oacp, controlplanev1alpha2.UpgradeCompletedCondition)
+		oacp.Status.Ready = true
+		if err := upgrader.VerifyUpgradedNodes(ctx, oacp); err != nil {
+			log.Error(err, "failed to verify control plane nodes after upgrade")
+			return ctrl.Result{}, err
+		}
 		return ctrl.Result{}, nil
 	}
 
--- a/controlplane/internal/upgrade/upgrade.go
+++ b/controlplane/internal/upgrade/upgrade.go
@@ -6,13 +6,16 @@ import (
 	"slices"
 	"strings"
 
+	controlplanev1alpha2 "github.com/openshift-assisted/cluster-api-agent/controlplane/api/v1alpha2"
 	"github.com/openshift-assisted/cluster-api-agent/controlplane/internal/release"
 	"github.com/openshift-assisted/cluster-api-agent/pkg/containers"
 
 	"github.com/openshift-assisted/cluster-api-agent/controlplane/internal/workloadclient"
 	configv1 "github.com/openshift/api/config/v1"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -36,6 +39,7 @@ type ClusterUpgrade interface {
 	GetCurrentVersion(ctx context.Context) (string, error)
 	IsDesiredVersionUpdated(ctx context.Context, desiredVersion string) (bool, error)
 	UpdateClusterVersionDesiredUpdate(ctx context.Context, desiredVersion string, options ...ClusterUpgradeOption) error
+	VerifyUpgradedNodes(ctx context.Context, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) error
 }
 
 func NewOpenshiftUpgradeFactory(remoteImage containers.RemoteImage, clientGenerator workloadclient.ClientGenerator) *OpenshiftUpgradeFactory {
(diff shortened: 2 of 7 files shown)


rccrdpccl:
Do we really want to set the ready status to false? the ready status should be there to inform about the controlplane readiness to receive requests, which should be unaffected during upgrades (and if it's affected, it shouldn't be solely because it's running an upgrade, but because some component is down).
https://cluster-api.sigs.k8s.io/developer/providers/contracts/control-plane#controlplane-initialization-completed

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -246,6 +247,7 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
 	oacp.Status.DistributionVersion, err = upgrader.GetCurrentVersion(ctx)
 	if err != nil {
 		log.V(logutil.WarningLevel).Info("failed to get OpenShift version from ClusterVersion", "error", err.Error())
+		return ctrl.Result{}, err
```
rccrdpccl, line 250: Why returning here? Even without current version we should be able to perform an upgrade

controlplane/internal/upgrade/upgrade.go
```diff
@@ -151,6 +155,34 @@ func (u *OpenshiftUpgrader) UpdateClusterVersionDesiredUpdate(ctx context.Contex
 	return nil
 }
 
+func (u *OpenshiftUpgrader) VerifyUpgradedNodes(ctx context.Context, oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane) error {
```
rccrdpccl, line 158: Shouldnt this be handled by https://github.com/openshift-assisted/cluster-api-agent/blob/92cebf02e87613e0194b807776e3abde5ecca98e/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go#L536C51-L536C70?

controlplane/internal/upgrade/upgrade.go
```diff
@@ -151,6 +155,34 @@ func (u *OpenshiftUpgrader) UpdateClusterVersionDesiredUpdate(ctx context.Contex
...
+	log := ctrl.LoggerFrom(ctx)
+	nodes := &corev1.NodeList{}
+	if err := u.client.List(ctx, nodes, client.MatchingLabels{"node-role.kubernetes.io/control-plane": ""}); err != nil {
+		return err
+	}
+	readyNodes := 0
+	for _, node := range nodes.Items {
+		// It's difficult to determine which OCP version is running on this Node, so we will
+		// just ensure that the Node is ready
+		for _, condition := range node.Status.Conditions {
```
rccrdpccl, line 168: we *might* want to add this logic for ready replicas, however this should not affect updated replicas
Updated replicas is a field that informs whether the specs are up to date or not.

Usually, this would inform about upgrades too, as one field would be Version. However we are using DistributionVersion which is not (yet) on the machine spec.

rccrdpccl:
as a generic remark, we should understand why this check is performed: is it making sure we are holding stuff during an upgrade? is it just for validation?
If we need to hold stuff, we need to make sure we hold it also while machine deployments machines are being updated, as the deployment would not have Version field, we are centralizing upgrade logic in the controlplane (even for worker nodes).

In any case I believe this should either be reflected in "UpdatedMachines" or some other conditions.

controlplane/internal/upgrade/upgrade.go
CrystalChun: Ideally it would be, but we can't actually update the machines so this a work-around

controlplane/internal/upgrade/upgrade.go
CrystalChun: This is a field that sylva checks for and if it's required for them to detect that upgrade finishes, then we should have it unless Jianzhu mentions otherwise

CrystalChun:
> Do we really want to set the ready status to false? the ready status should be there to inform about the controlplane readiness to receive requests, which should be unaffected during upgrades (and if it's affected, it shouldn't be solely because it's running an upgrade, but because some component is down). https://cluster-api.sigs.k8s.io/developer/providers/contracts/control-plane#controlplane-initialization-completed

This is a field that Sylva checks that upgrade has started and is finished (once set to true again). Should we have Sylva change their checks?

rccrdpccl:
> This is a field that Sylva checks that upgrade has started and is finished (once set to true again). Should we have Sylva change their checks?

On upgrades (especially rolling upgrades) it could happen that a node is restarting some components and the cluster controlplane would become unavailable. Checking ready makes sense, however we should report the real status: if the controlplane is ready, we should be reporting ready. IMO we should not fake the ready status to report upgrade in progress.

controlplane/internal/upgrade/upgrade.go
rccrdpccl: I mean that even if we want to run the workaround, I think we should keep the logic updating those fields together (we should make sure it runs that part of reconciliation during upgrades, but still I don't think we should split the logic handling those fields)

CrystalChun:
Riccardo led a discussion offline with Jianzhu and I.

Jianzhu will ask Sylva if they're ok with creating a separate checking unit for openshift upgrades which will verify if the control plane's status upgrade completed condition is true to signal a completed upgrade. 

We'll wait for that decision, and if they do not accept it, then we will need to pivot and either continue this hacking path or find a different solution that can satisfy both the CAPI standards and Sylva's upgrade expectations.

jianzzha:
The Sylva discussion: https://gitlab.com/sylva-projects/sylva-core/-/issues/2191#note_2390882900

Please check and make sure what I said regarding the "status" in this thread accurately reflected what we discussed. I also have this concern that OpenShift behaves differently than other k8s distros used in Sylva. I understand the reasoning behind it but we do need to avoid that as much as possible.

controlplane/internal/upgrade/upgrade.go
jianzzha: Even though machine is immutable, their status can change, right? I assume for MD, if its owned machine status.conditions.Ready changes to false during the upgrade, the MD status will be updated to reflect that right? Same logic for controlplane machine? 

Does openshift upgrade provide the granular to tell which node is being updated @CrystalChun

jianzzha: I guess what I'm trying to say is that the agent capi provider control is aware of the upgrade and should change the machine status, not the core CAPI provider.

jianzzha: the new `MachinesUpToDate` for MD and controlplane in v1beta2 might work?  https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240916-improve-status-in-CAPI-resources.md

jianzzha: The `UpToDate` semantics seems to change with this inplace upgrade proposal:  https://github.com/kubernetes-sigs/cluster-api/blob/ad3312b627195d43434946ae7557fd12985ed084/docs/proposals/20240807-in-place-updates.md#high-level-flow

It sets the machine  `UpToDate`  condition to false when the upgrade begins and set it to true when the upgrade is complete. I'm guessing this need to work together with `pending-hooks: ExternalUpdate` annotation to have the "upgrade" semantics.

I think we need to consider if we have a 'CAPI compliance' path. Right now we might be in a non-compliance position for upgrade: the MD owned machines are being upgraded but the MD and machine controllers have no visibility to what's going on therefore can not update status.  When the distributionversion is accepted by upstream, I think the situation won't change because we still can not have the distributionversion under MD, otherwise changing distributionversion under MD would trigger MD to execute rolling upgrade, right?

For openshift inplace upgrade to be 'CAPI compliance', I guess we will need an "openshift friendly inplace upgrade scheme" accepted by upstream.

rccrdpccl: > The UpToDate semantics seems to change with this inplace upgrade proposal: https://github.com/kubernetes-sigs/cluster-api/blob/ad3312b627195d43434946ae7557fd12985ed084/docs/proposals/20240807-in-place-updates.md#high-level-flow

Are you suggesting we should already anticipate that? I think we might get around the issue of the owner changing back the status by pausing the machines for the upgrade duration. I think we can explore this path

>  Right now we might be in a non-compliance position for upgrade
Yes, until both distributionVersion and in-place-updates are implemented, we are not 100% complying with the CAPI contract/way of doing things

> I think the situation won't change because we still can not have the distributionversion under MD, otherwise changing distributionversion under MD would trigger MD to execute rolling upgrade, right?

We cannot have distributionVersion under MD because we do not control that API at the moment. And yes, even if we could, not supporting in-place upgrade changing such field will trigger a rolling upgrade.

rccrdpccl:
@CrystalChun can this be closed?

CrystalChun:
It looks like Sylva's not accepting just the condition that upgrade is completed yet. I'll close this for now then, but I'm worried if they don't accept it then we may need to do something else like this 😅

PR #94, opened 2025-03-10
MGMT-20101: Allow additional CA bundle for assisted ignition from configmap

carbonin:
At startup the bootstrap provider will create the assisted http client using both the referenced configmap (if any) and the system CA bundle

This also requires RBAC for reading configmaps at a cluster level

Tested this by deploying a cluster in OpenShift using provider images built from this branch by setting the new configmap envvars in the bootstrap provider components:
```
        - name: ASSISTED_CA_BUNDLE_NAME
          value: default-ingress-cert
        - name: ASSISTED_CA_BUNDLE_NAMESPACE
          value: openshift-config-managed
```

Resolves https://issues.redhat.com/browse/MGMT-20101


bootstrap/main.go
```diff
@@ -193,3 +210,44 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func getAssistedHTTPClient(config assistedinstaller.ServiceConfig, c client.Client) (*http.Client, error) {
```
rccrdpccl, line 214: I think this should go in [this](https://github.com/openshift-assisted/cluster-api-agent/tree/master/assistedinstaller) package

PR #95, opened 2025-03-13
NO-ISSUE: Add ansible plays for e2e tests

CrystalChun:
Separates the main play into the following ansible plays for easily running one section of the test:
- cleanup.yaml: cleans up the environment by removing all hosts, vms and the kind cluster
- deploy_controllers.yaml: allows for redeploying the controllers in an already setup test environment. Useful for testing code changes
- hosts.yaml: Creates the host VMs and BMHs

## Summary by CodeRabbit

- **New Features**
	- Introduced automation playbooks to streamline test environment setup, including controller deployment and host configuration.
	- Added a cleanup process to gracefully dismantle test environments.
- **Documentation**
	- Included detailed instructions and usage examples for the new automation tasks to assist with deployments and cleanups.

### DIFF
--- a/test/ansible/README.md
+++ b/test/ansible/README.md
@@ -1,0 +1,31 @@
+# Additional Ansible Plays
+
+## Deploy controllers
+
+`deploy_controllers.yaml` allows for redeploying the OpenShift Agent Control Plane and OpenShift Agent Bootstrap controllers in an already setup test environment. Useful for testing code changes live.
+
+### Usage
+
+```sh
+# Set a container tag that's different from the current deployed tag
+export CONTAINER_TAG=
+
+# Recreate the manifests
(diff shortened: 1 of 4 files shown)


rccrdpccl:
Nice idea, however can we extract steps in reusable roles? That way if we change one, we do not need to propagate the change in multiple files.

No need to do all at once, if we can split into reusable pieces we can tackle bit by bit, but let's try not to introduce duplicate tasks

CrystalChun:
Definitely! This was just a first pass that I did so I can perform these easily and I thought I could push it up so we both can use it, but will come back to this to create reusable roles!

rccrdpccl:
@CrystalChun can this be closed? The e2e tests have been now refactored, it would be much easier to create different playbooks.

PR #96, opened 2025-03-14
MGMT-20013: Allow user-specified capabilities when installing a workload cluster

CrystalChun:
The `additionalEnabledCapabilities` spec field on the `OpenshiftAssistedControlPlane` allows users to specify any additional capabilities they want enabled when installing a workload cluster.

---
Internally, this will set the `InstallConfigOverride` annotation on the `AgentClusterInstall` when it's generated. 

- For SNO (single node openshift) and non-baremetal platform MNO (multi-node openshift), this is set to the `OpenshiftAssistedControlPlane` spec field or left empty if the spec field is empty
- For MNO (multi-node openshift), this is set to the default list for baremetal platform, but the `OpenshiftAssistedControlPlane` spec field can specify additional capabilities to be appended to this list.
    - It will ignore duplicates (e.g. if `baremetal` is specified in the spec field, it will not appear twice in the annotatioin)
    - It will ignore `MachineAPI` if it's specified
    - Default list: `baremetal`, `Console`, `Insights`, `OperatorLifecycleManager`, `Ingress`

---
Example usage:
```yaml
kind: OpenshiftAssistedControlPlane
metadata:
  name: test-multinode
spec:
  config:
    apiVIPs:
    - 192.168.222.40
    ingressVIPs:
    - 192.168.222.41
    additionalEnabledCapabilities:
    - "NodeTuning"
```

---
Example output:
```yaml
  kind: AgentClusterInstall
  metadata:
    annotations:
      agent-install.openshift.io/install-config-overrides: '{"capabilities": {"baselineCapabilitySet":
        "None", "additionalEnabledCapabilities": ["baremetal","Console","Insights","OperatorLifecycleManager","Ingress","NodeTuning"]}}"'
```
 

## Summary by CodeRabbit

- **New Features**
	- Introduced advanced configuration options that let users specify a default set and additional installation capabilities.
	- Enhanced the installation process with dynamic, platform-specific defaults for both bare metal and non-bare metal environments, enabling a more flexible and reliable cluster setup.
	- Added a utility function for initializing `OpenshiftAssistedControlPlane` with specified capabilities.

- **Bug Fixes**
	- Improved handling of capabilities to prevent duplicates and exclusions during installation.

- **Tests**
	- Expanded test coverage for capability handling in the `ClusterDeployment Controller`, validating various scenarios for both bare metal and non-bare metal clusters.


controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +281,41 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+	isBaremetalPlatform := aci.Spec.PlatformType == hiveext.BareMetalPlatformType
+
+	additionalCapabilitiesList := []string{}
+	if !isSNO && isBaremetalPlatform {
+		additionalCapabilitiesList = append([]string{}, defaultMNOAdditionalCapabilities...)
+	}
+
+	for _, capability := range oacp.Spec.Config.AdditionalEnabledCapabilities {
+		// Ignore MAPI for baremetal MNO clusters and ignore duplicates
+		if (strings.EqualFold(capability, "MachineAPI") && !isSNO && isBaremetalPlatform) || funk.ContainsString(additionalCapabilitiesList, capability) {
```
rccrdpccl, line 302: I think we do not really care if this is SNO or not, right? we only care about the platform? Is MAPI capability added only for baremetal platform?

controlplane/config/crd/bases/controlplane.cluster.x-k8s.io_openshiftassistedcontrolplanes.yaml
```diff
@@ -92,6 +92,13 @@ spec:
               config:
                 description: Config specs for the OpenshiftAssistedControlPlane
                 properties:
+                  additionalEnabledCapabilities:
```
rccrdpccl, line 95: How about using the same interface as config install? https://github.com/openshift/installer/blob/main/pkg/types/installconfig.go#L550

I think this would still meet requirement plus give us greater flexibility, as in the end this is what we are going to push through. The only thing we need to validate is to make sure we do not turn on MAPI.

controlplane/internal/controller/clusterdeployment_controller_test.go
```diff
@@ -188,4 +188,186 @@ var _ = Describe("ClusterDeployment Controller", func() {
...
+						Replicas: int32(3),
+						Config: v1alpha2.OpenshiftAssistedControlPlaneConfigSpec{
+							AdditionalEnabledCapabilities: []string{"baremetal", "MachineAPI", "NodeTuning"},
+						},
+					},
+				}
+				aci := &hiveext.AgentClusterInstall{}
+				expectedAnnotation := `{"capabilities": {"baselineCapabilitySet": "None", "additionalEnabledCapabilities": ["baremetal","MachineAPI","NodeTuning"]}}"`
+
+				err := setACICapabilities(oacp, aci)
```
rccrdpccl, line 252: In general, it'd be best not testing private methods. We should provide input and test the expected output (i.e. oacp specs, and expect the ACI to have this annotation) without directly calling this private method

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: > I think we do not really care if this is SNO or not, right? we only care about the platform?

Oh yes good catch 😆 

> Is MAPI capability added only for baremetal platform?

MAPI should not be added for baremetal platform

controlplane/config/crd/bases/controlplane.cluster.x-k8s.io_openshiftassistedcontrolplanes.yaml
CrystalChun: Good idea, thanks! Will update

controlplane/internal/controller/clusterdeployment_controller_test.go
CrystalChun: Ah right, sorry I forgot about not testing private methods. Will update. Thanks!

CrystalChun: Will fully update later, but keeping these for now to verify the function does what it's supposed to.

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +284,82 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+		if isBaremetalPlatform {
+			baselineCapability = defaultBaremetalBaselineCapability
+		}
+		aci.Annotations = map[string]string{
+			InstallConfigOverrides: fmt.Sprintf(`{"capabilities": {"baselineCapabilitySet": "%s", "additionalEnabledCapabilities": %s}}"`, baselineCapability, string(additionalEnabledCapabilities)),
+		}
+		return nil
+	}
+
+	// Otherwise only baseline capability is set
```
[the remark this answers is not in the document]
CrystalChun: Yes that would be set before we get here if both are customized by the user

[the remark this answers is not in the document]
rccrdpccl: I might be misunderstanding, but I can only see 2 places where we edit annotations: the block above (L316) is guarded by `baselineCapability == ""`, the block below does not specify additionalEnabledCapabilities.
What happens if somebody wants some totally custom capabilities, as we are doing for baremetal platform. Can this be done? baselineCapabilitySet: none, additionalCapabilities: ["foo","bar",....]
I could not find the test for this case, could it be added?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +284,82 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+	if additionalEnabledCapabilities == "" && baselineCapability == "" {
+		return nil
+	}
+
+	if baselineCapability == "" {
+		// Only AdditionalEnabledCapabilities are set so we should set the default baseline capability
+		baselineCapability = defaultBaselineCapability
+		if isBaremetalPlatform {
+			baselineCapability = defaultBaremetalBaselineCapability
+		}
```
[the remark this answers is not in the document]
rccrdpccl: I think it's good not setting the annotation if not necessary, if `oacp.Spec.Config.Capabilities == Capabilities{}` it means it's not set by the user, right?  Otherwise we can use the smart baselineCapability value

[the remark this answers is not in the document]
CrystalChun: Yes, but we still need to detect if it's platform baremetal and set default capabilities even if they're not set by the user.

controlplane/internal/controller/clusterdeployment_controller.go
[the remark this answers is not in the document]
CrystalChun: Oh 🤦‍♀️ good catch, I totally missed the case where both are set. Thank you! Will also add a test for this

controlplane/internal/controller/clusterdeployment_controller.go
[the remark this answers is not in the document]
CrystalChun: Ok modified it so that we don't set annotations if it's not baremetal and there's no capabilities specified. What do you think?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +284,85 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+
+// setACICapabilities will set the install config override annotation in the AgentClusterInstall if there
+// are additional enabled capabilities that need to be defined.
+// For SNO (single node openshift) and non-baremetal platform MNO (multi-node openshift), this is set if
+// the OpenshiftAssistedControlPlane has additional capabilities set.
+// For MNO (multi-node openshift), this is set to the default list for baremetal platform,
+// but the OpenshiftAssistedControlPlane can specify additional capabilities to be appended to this list.
+func setACICapabilities(oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane, aci *hiveext.AgentClusterInstall) error {
+	isBaremetalPlatform := aci.Spec.PlatformType == hiveext.BareMetalPlatformType
+	if funk.IsEmpty(oacp.Spec.Config.Capabilities) && !isBaremetalPlatform {
```
rccrdpccl, line 296: I think we shouldn't add the `funk` dependency just to check for zero value.

if we want to make it more readable, we can do something
```
if !areCapabilitiesSet(oacp) {
   return nil
}
....
```

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +284,85 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+	additionalEnabledCapabilities, err := getAdditionalCapabilities(oacp.Spec.Config.Capabilities.AdditionalEnabledCapabilities, isBaremetalPlatform)
+	if err != nil {
+		return err
+	}
+
+	installConfigOverride := fmt.Sprintf(`{"capabilities": {"baselineCapabilitySet": "%s", "additionalEnabledCapabilities": %s}}`, baselineCapability, additionalEnabledCapabilities)
+	if additionalEnabledCapabilities == "" {
+		// The user only specified BaselineCapability so no need to set additionalEnabledCapabilities in the override
+		installConfigOverride = fmt.Sprintf(`{"capabilities": {"baselineCapabilitySet": "%s"}}`, baselineCapability)
+	}
```
rccrdpccl, line 313: nit: if we'd have an output struct to unmarshal capabilities in, the logic could be slightly simpler (omitempty for the additionalEnableCapabilities, no sprintfs, no conditional here)

controlplane/internal/controller/clusterdeployment_controller.go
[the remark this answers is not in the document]
rccrdpccl: nice :heart_eyes:

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Good idea, tried to do it with a new type, wdyt?

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Good idea. Thanks! Added it in, what do you think?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +285,92 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
 	return nil
 }
+
+type InstallConfigOverride struct {
+	Capability Capability `json:"capabilities,omitempty"`
+}
+type Capability struct {
+	BaselineCapability            string   `json:"baselineCapabilitySet,omitempty"`
+	AdditionalEnabledCapabilities []string `json:"additionalEnabledCapabilities,omitempty"`
+}
```
rccrdpccl, line 295: Isn't this exactly the same as https://github.com/openshift/installer/blob/74c0c30aa1e7926302b07564480a8eb2f605be20/pkg/types/installconfig.go#L551 ? Shouldn't we use that instead?
It's ok for the OpenshiftAssistedControlPlane spec, because we might be adding additional validation and we can have better control of our API, but this is intended for install config so I think it would make more sense. Would you agree?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +285,92 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+
+	if len(additionalCapabilitiesList) < 1 {
+		return nil
+	}
+
+	return additionalCapabilitiesList
+}
+
+func capabilitiesSet(capabilities v1alpha2.Capabilities) bool {
+	return capabilities.BaselineCapability != "" || len(capabilities.AdditionalEnabledCapabilities) > 0
```
rccrdpccl, line 375: Wouldn't comparing zero value for Capabilities also work?

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -273,3 +285,92 @@ func (r *ClusterDeploymentReconciler) createImageRegistry(ctx context.Context, r
...
+
+func getAdditionalCapabilities(specifiedAdditionalCapabilities []string, isBaremetalPlatform bool) []string {
+	additionalCapabilitiesList := []string{}
+	if isBaremetalPlatform {
+		additionalCapabilitiesList = append([]string{}, defaultBaremetalAdditionalCapabilities...)
+	}
+
+	for _, capability := range specifiedAdditionalCapabilities {
+		// Ignore MAPI for baremetal MNO clusters and ignore duplicates
+		if (strings.EqualFold(capability, "MachineAPI") && isBaremetalPlatform) || funk.ContainsString(additionalCapabilitiesList, capability) {
```
rccrdpccl, line 361: `slices.Contains` is preferable as it doesn't add a dependency on `funk`

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Screenshot 2025-03-18 at 10 53 45 AM

Hmm I get this when I try to compare the empty struct... Do you know of another way to compare the zero value?

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Sure changed it

controlplane/internal/controller/clusterdeployment_controller.go
CrystalChun: Changed it, thanks!

controlplane/internal/controller/clusterdeployment_controller.go
rccrdpccl: Oh, right, I didn't realize, sorry :sweat: 

In any case I think we can solve it with apimachinery equality package. which we already use in the project:
```equality.Semantic.DeepEqual(myCapabilities, zeroCapabilities)```

CrystalChun: Oh no worries! Thank you for recommendation, I modified it to use the equality package

rccrdpccl:
nice work!

/lgtm
/approve

PR #102, opened 2025-03-19
MGMT-20067: remove arch suffix from Spec.DistributionVersion

rccrdpccl:
Architecture should be fetched from bootstrap config instead.
If we have the same arch for all nodes, we should use arch-specific image, multi otherwise.

## Summary by CodeRabbit

- **New Features**
  - Enhanced release image retrieval to support multi-architecture configurations during upgrades, ensuring that the appropriate release image is selected based on the target CPU architectures.
  - Introduced new utility functions for creating `OpenshiftAssistedConfig` objects with and without references to `InfraEnv`.

- **Refactor**
  - Streamlined version strings in configuration examples and playbooks by removing redundant architecture suffixes, resulting in more consistent and simplified version declarations.

- **Bug Fixes**
  - Improved test coverage for architecture-specific configurations in the ClusterDeployment Controller tests.

### DIFF
--- a/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
+++ b/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
@@ -20,11 +20,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	semver "github.com/blang/semver/v4"
-	"github.com/go-logr/logr"
 
 	"github.com/openshift-assisted/cluster-api-agent/assistedinstaller"
 	bootstrapv1alpha1 "github.com/openshift-assisted/cluster-api-agent/bootstrap/api/v1alpha1"
@@ -191,7 +189,11 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	releaseImage := getReleaseImage(*oacp)
+	architecture, err := getArchitectureFromBootstrapConfigs(ctx, r.Client, oacp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	releaseImage := getReleaseImage(*oacp, architecture)
 
 	k8sVersion, err := r.K8sVersionDetector.GetKubernetesVersion(releaseImage, string(pullsecret))
 	markKubernetesVersionCondition(oacp, err)
@@ -210,7 +212,7 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
 	result := ctrl.Result{}
 	if conditions.IsTrue(oacp, controlplanev1alpha2.KubeconfigAvailableCondition) {
 		// in case upgrade is still in progress, we want to requeue, however we also want to reconcile replicas
-		result, err = r.upgradeWorkloadCluster(ctx, cluster, oacp, log, pullsecret)
+		result, err = r.upgradeWorkloadCluster(ctx, cluster, oacp, architecture, pullsecret)
 		if err != nil {
 			return result, err
 		}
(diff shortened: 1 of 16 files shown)


controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -210,15 +212,47 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
...
+	if err := k8sClient.List(ctx, &configList, listOptions...); err != nil {
+		return []string{}, err
+	}
+
+	architectures := make([]string, 0)
+	for _, config := range configList.Items {
+		if config.Spec.CpuArchitecture != "" {
+			architectures = append(architectures, config.Spec.CpuArchitecture)
+		}
+	}
```
CrystalChun, line 249: What do you think about moving the logic from [`getArchitecture`](https://github.com/openshift-assisted/cluster-api-agent/pull/102/files#diff-9bf377e2c7bc277e1ec29fb657fa201dacbe0e599fa8abea4cc1b577425021c6R66-R83) to here so that way we only need to loop through once to get the architecture?

test/utils/unit.go
```diff
@@ -237,3 +238,33 @@ func NewBareMetalHost(namespace, name string) *v1alpha1.BareMetalHost {
...
+	namespace, name, clusterName string,
+	infraEnv *v1beta1.InfraEnv,
+) *v1alpha2.OpenshiftAssistedConfig {
+	var ref *corev1.ObjectReference
+	if infraEnv != nil {
+		ref = &corev1.ObjectReference{
+			Namespace: infraEnv.GetNamespace(),
+			Name:      infraEnv.GetName(),
+		}
+	}
```
CrystalChun, line 252: Do we need this check? Couldn't we just set the InfraEnvRef to whatever's passed in even if it's nil?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
rccrdpccl: Good idea! The code is actually more clear this way. I've decided to loop through them twice for simplicity and better readability. I don't think it's problematic performance wise, as the List operation is the most expensive one, second pass it's only in memory architecture strings comparison. WDYT?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -210,15 +212,71 @@ func (r *OpenshiftAssistedControlPlaneReconciler) Reconcile(ctx context.Context,
...
+		client.MatchingLabels(labelSelector),
+	}
+	var configList bootstrapv1alpha1.OpenshiftAssistedConfigList
+	if err := k8sClient.List(ctx, &configList, listOptions...); err != nil {
+		return "", err
+	}
+
+	architectures := make([]string, 0)
+	for _, config := range configList.Items {
+		architectures = append(architectures, config.Spec.CpuArchitecture)
```
CrystalChun, line 250: Sorry, I couldn't find in the codebase where this is set 😅 Where is `config.Spec.CpuArchitecture` set?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
CrystalChun: Looks great!! 🤩

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
rccrdpccl: It's never set, it has a default value https://github.com/openshift-assisted/cluster-api-agent/blob/master/bootstrap/api/v1alpha1/openshiftassistedconfig_types.go#L58

In any case, if it's empty, we consider them to be `multi` arch

rccrdpccl:
/hold

holding to allow release without this change

PR #103, opened 2025-03-19
MGMT-20162: improve usage of CreateOrUpdate

rccrdpccl:
The usage of CreateOrUpdate in this project is mainly to avoid boilerplate when running both operations.
We do not change any value depending on the operation.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined deployment and update workflows by removing redundant methods and simplifying error handling.
- **New Features**
  - Introduced a new utility function for creating or updating Kubernetes objects, enhancing efficiency in resource management.
  - Added a new file for utility functions to support the updated workflows.
  - Added a test suite for utility functions to ensure reliability and correctness in deployment operations.

### DIFF
--- a/controlplane/internal/controller/utils.go
+++ b/controlplane/internal/controller/utils.go
@@ -1,0 +1,1 @@
+package controller
--- a/controlplane/internal/controller/clusterdeployment_controller.go
+++ b/controlplane/internal/controller/clusterdeployment_controller.go
@@ -37,7 +37,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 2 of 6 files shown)


controlplane/internal/controller/utils.go
```diff
@@ -0,0 +1,24 @@
+package controller
+
+import (
+	"context"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NOOP mutate function. Useful for CreateOrUpdate when the only thing we want to do is create or upgrade
```
CrystalChun, line 10:
```suggestion
// NOOP mutate function. Useful for CreateOrUpdate when the only thing we want to do is create or update
```

controlplane/internal/controller/utils.go
```diff
@@ -0,0 +1,24 @@
...
+// the object without any change depending on the operation performed
+func mutateNOOP() error {
+	return nil
+}
+
+// Create or upgrade object. Ignore AlreadyExists error
```
CrystalChun, line 16:
```suggestion
// Create or update object. Ignore AlreadyExists error
```

rccrdpccl:
> Very minor wording thing, but I think this is a great idea/change!
> 
> One thing I'm not sure of is if the update part of the function actually updates the object. I noticed when I was testing other things that when the `createOrUpdate` function was run on an existing CR, it wouldn't really update it. It could be because of my own testing flake, but just wanted to point this out so we can ensure it actually updates the object!

Nice catch, I actually was looking into that and did not see it... :facepalm: 
Indeed it wasn't updating the object https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.3/pkg/controller/controllerutil/controllerutil.go#L307-L335

It should be working now

CrystalChun (approved):
/lgtm

Very nice!!

CrystalChun (approved):
/lgtm

PR #121, opened 2025-04-02
Use controllerutil for create or update

carbonin:
The current create or update function causes problems when called for an update with a resource that was freshly created in code (not queried from the existing cluster).

This is solved by using `controllerutil.CreateOrPatch` which handles doing this query, applying changes, and only updating if changes were made.

This PR updates all our uses of the existing util function to the `controllerutil` version as well as cleans up some existing functions to _ensure_ :wink: they're only doing one thing.

----------------------------------------------

WIP until I take care of the last caller and remove the function.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the cluster management process with improved logic for managing cluster deployment references and configurations.
  - Updated image registry configuration generation to return simplified data outputs.
- **Refactor**
  - Streamlined the reconciliation process for cluster deployments and image registry configurations.
- **Bug Fixes**
  - Addressed issues related to error handling in the cluster deployment logic.
- **Tests**
  - Adjusted test cases to align with the updated function names and logic for image registry data generation.
- **Chores**
  - Removed outdated utility functions and test files to clean up the codebase.


rccrdpccl:
I finally understand why controlleruitls' CreateOrUpdate it's so inconvenient to use.. :sweat: 

We should probably remove our CreateOrUpdate within this PR too

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -500,30 +508,73 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensureClusterDeployment(
 	acp *controlplanev1alpha2.OpenshiftAssistedControlPlane,
 	clusterName string,
 ) error {
-	if acp.Status.ClusterDeploymentRef == nil {
-		clusterDeployment := assistedinstaller.GetClusterDeploymentFromConfig(acp, clusterName)
-		_ = controllerutil.SetOwnerReference(acp, clusterDeployment, r.Scheme)
-		if err := util.CreateOrUpdate(ctx, r.Client, clusterDeployment); err != nil {
+	if acp.Status.ClusterDeploymentRef != nil {
+		return nil
+	}
```
CrystalChun, line 513: Wouldn't we still want to go through the create or patch function below even if the cluster deployment ref is set in the ACP's status? Like in the event the ACP's pull secret ref change?

CrystalChun:
One question, but looks great otherwise!

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
carbonin: I agree generally, but this was the behavior before so I just kept it as is for this PR.

carbonin:
/hold

Want to fix one more thing

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
CrystalChun: Gotcha, sorry didn't catch that it was prior behavior.

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -88,44 +89,94 @@ func (r *ClusterDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		aci.Spec.PlatformType = hiveext.PlatformType(configv1.NonePlatformType)
+		aci.Spec.ProvisionRequirements = hiveext.ProvisionRequirements{
+			ControlPlaneAgents: int(oacp.Spec.Replicas),
+			WorkerAgents:       workerNodes,
+		}
+		aci.Spec.DiskEncryption = oacp.Spec.Config.DiskEncryption
+		aci.Spec.MastersSchedulable = oacp.Spec.Config.MastersSchedulable
+		aci.Spec.Proxy = oacp.Spec.Config.Proxy
+		aci.Spec.SSHPublicKey = oacp.Spec.Config.SSHAuthorizedKey
+		aci.Spec.ImageSetRef = &hivev1.ClusterImageSetReference{Name: clusterDeployment.Name}
```
CrystalChun, line 158: I think the ClusterImageSet name is based off of the OACP spec 

```suggestion
                 if oacp.Spec.Config.ImageRegistryRef != nil {
		    aci.Spec.ImageSetRef = &hivev1.ClusterImageSetReference{Name: acp.Spec.Config.ImageRegistryRef.Name}
		}
```

carbonin: hmmm maybe it should be but I'm pretty sure what I did here is the same as what we started with.

You can see it a bit more easily if you look at each commit, but I know there ended up being a lot of them :sweat_smile: 

In current master we pass `clusterDeployment.Name` into `computeClusterImageSet` which uses it directly as a parameter to name the image set.

carbonin: And to take it further `computeAgentClusterInstall` takes in the `imageSet` object which is used for its name.

CrystalChun: Oh 🤦‍♀️ I was looking at the wrong function, sorry! That's right it is the cluster deployment name, sorry ignore this 😓

CrystalChun (approved):
/lgtm

PR #126, opened 2025-04-08
MGMT-19979: Propagate ClusterVersion upgrade status message

CrystalChun:
Adds the upgrade message from the ClusterVersion in the workload cluster to the OACP's upgrade status so it's transparent to the user. Also adds a failure reason in case the upgrade fails.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new constant for upgrade failure reasons, enhancing clarity on upgrade issues.
  - Improved upgrade status feedback with detailed messages for ongoing and failed upgrades.

- **Tests**
  - Expanded test coverage validates the improved upgrade status handling and message consistency under various upgrade scenarios.

### DIFF
--- a/controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
+++ b/controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
@@ -81,6 +81,7 @@ var _ = Describe("OpenshiftAssistedControlPlane Controller", func() {
 			mockUpgrader.EXPECT().IsUpgradeInProgress(gomock.Any()).Return(false, nil).AnyTimes()
 			mockUpgrader.EXPECT().GetCurrentVersion(gomock.Any()).Return("4.18.0", nil).AnyTimes()
 			mockUpgrader.EXPECT().IsDesiredVersionUpdated(gomock.Any(), gomock.Any()).Return(true, nil).AnyTimes()
+			mockUpgrader.EXPECT().GetUpgradeStatus(gomock.Any()).Return("", nil).AnyTimes()
 
 			mockUpgradeFactory = upgrade.NewMockClusterUpgradeFactory(ctrl)
 			mockUpgradeFactory.EXPECT().NewUpgrader(gomock.Any()).Return(mockUpgrader, nil).AnyTimes()
@@ -384,6 +385,7 @@ var _ = Describe("Upgrade scenarios", func() {
 		mockUpgradeFactory.EXPECT().NewUpgrader(gomock.Any()).Return(mockUpgrader, nil)
 		mockUpgrader.EXPECT().IsUpgradeInProgress(gomock.Any()).Return(false, nil)
 		mockUpgrader.EXPECT().GetCurrentVersion(gomock.Any()).Return(currentVersion, nil)
+		mockUpgrader.EXPECT().GetUpgradeStatus(gomock.Any()).Return("", nil)
 		mockUpgrader.EXPECT().IsDesiredVersionUpdated(gomock.Any(), desiredVersion).Return(false, nil)
 		mockUpgrader.EXPECT().UpdateClusterVersionDesiredUpdate(gomock.Any(), desiredVersion, gomock.Any(), gomock.Any()).Return(nil)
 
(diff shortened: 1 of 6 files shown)


controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
```diff
@@ -506,8 +517,9 @@ var _ = Describe("Upgrade scenarios", func() {
 		Expect(k8sClient.Update(ctx, openshiftAssistedControlPlane)).To(Succeed())
 
 		mockUpgradeFactory.EXPECT().NewUpgrader(gomock.Any()).Return(mockUpgrader, nil)
-		mockUpgrader.EXPECT().IsUpgradeInProgress(gomock.Any()).Return(false, nil)
+		mockUpgrader.EXPECT().IsUpgradeInProgress(gomock.Any()).Return(true, nil) // should this be successfully starting upgrade?
```
CrystalChun, line 520: Hey @rccrdpccl should this test `should handle upgrade with repository override` return true that upgrade is in progress? I wasn't sure

CrystalChun:
Thanks @rccrdpccl 🙏 added checks to existing tests for correct statuses. Let me know if more tests are needed though!

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
rccrdpccl: According to the current behaviour, this should start an upgrade either way:
* if UpgradeInProgress, but DesiredVersion is not updated, we should trigger an update
* if Upgrade not in progress, and desiredVersion is not updated, we should still trigger an update.

I think the logic is fair, but would this work with CVO? Can we update the version if an upgrade is already running?

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
```diff
@@ -528,6 +540,7 @@ var _ = Describe("Upgrade scenarios", func() {
 		condition := conditions.Get(openshiftAssistedControlPlane, controlplanev1alpha2.UpgradeCompletedCondition)
 		Expect(condition).NotTo(BeNil())
 		Expect(condition.Status).To(Equal(corev1.ConditionFalse))
+		Expect(condition.Message).To(Equal("upgrade to version 4.15.0 in progress\nupgrading cluster"))
```
rccrdpccl, line 543: Do we need this message? I think it's superflous (the `upgrading cluster` part)

CrystalChun: The `upgrading cluster` message is just a stand-in for what CVO would report back and it's going to be appended to the final message we show in our status. Do you mean I should change the message to be more relevant?

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
CrystalChun: Good question, I can verify that!

controlplane/internal/controller/openshiftassistedcontrolplane_controller_test.go
rccrdpccl: I mean we can omit it if everything is OK: as we already say "upgrade to version X in progress"  I think there's no need to append "upgrading cluster"
In case of failure, I think adding further details is great

CrystalChun: Gotcha! That makes sense to show only relevant information and especially not repeating things. Thanks! Removed the upgrading cluster part of the message.

controlplane/internal/upgrade/upgrade.go
```diff
@@ -113,6 +114,21 @@ func (u *OpenshiftUpgrader) GetCurrentVersion(ctx context.Context) (string, erro
...
+	if err != nil {
+		return "", err
+	}
+	for _, condition := range clusterVersion.Status.Conditions {
+		if condition.Status == configv1.ConditionTrue {
+			return condition.Message, nil
+		}
+	}
+	return "", nil
+}
```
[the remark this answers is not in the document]
rccrdpccl: @CrystalChun is the rabbit's concern real?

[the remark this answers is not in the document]
rccrdpccl: So in this logic we're picking up the first message of any of those conditions and report that? Is that because they are sorted from most recent to least recent?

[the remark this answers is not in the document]
CrystalChun: It sounds like these are mutually exclusive, so the logic picks up any condition that's marked as true at the time.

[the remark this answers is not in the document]
rccrdpccl: Can we please add a little comment about this so the next person reading this code wouldn't be confused about it? :pray:

PR #135, opened 2025-04-14
handle CA bundle for k8s platform

rccrdpccl:
Turns out assisted-installer cert on k8s platform it's self signed, we were testing with an old infra-operator image that wouldn't enforce https.


rccrdpccl:
/retest

timeout

PR #137, opened 2025-04-23
defaults CA bundle references to kubernetes default values

rccrdpccl:
Reflect kubernetes default values for the CA bundle location, as MCE is already overriding some values


carbonin:
When you say "MCE is already overriding some values" I ssume you mean the the helm chart [here](https://github.com/stolostron/cluster-api-installer/blob/main/charts/cluster-api-provider-openshift-assisted/templates/apps_v1_deployment_capoa-bootstrap-controller-manager.yaml#L26-L29)? If so, we should be sure to also override the key and resource before merging this because right now I think those are relying on the old defaults.

rccrdpccl:
> When you say "MCE is already overriding some values" I ssume you mean the the helm chart [here](https://github.com/stolostron/cluster-api-installer/blob/main/charts/cluster-api-provider-openshift-assisted/templates/apps_v1_deployment_capoa-bootstrap-controller-manager.yaml#L26-L29)? If so, we should be sure to also override the key and resource before merging this because right now I think those are relying on the old defaults.

Good point:
https://github.com/stolostron/cluster-api-installer/pull/64
https://github.com/stolostron/cluster-api-installer/pull/63

carbonin (approved):

PR #148, opened 2025-05-02
Fix bootstrap konflux

carbonin:
Additional commit on top of https://github.com/openshift-assisted/cluster-api-provider-openshift-assisted/pull/142 to satisfy the enterprise contract requirements


carbonin:
Just realized I have write permissions here so I'm going to push to the existing PR instead of creating a new one

PR #165, opened 2025-05-06
Add renovate.json to prevent dependency updates

carbonin:
We don't want go module updates in the 2.9 branch.
These will be updated as needed to master and backported or fast-forwarded to this branch.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,7 @@
+{
+    "schedule": ["on Saturday"],
+    "groupName": "Konflux build pipeline",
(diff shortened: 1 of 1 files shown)


carbonin:
TIL I need to go into the GH actions/checks UI to retest the security scan :tired_face:

carbonin:
So it seems like this didn't work ... not sure why as the [IBIO one](https://github.com/openshift/image-based-install-operator/blob/backplane-2.9/renovate.json) is the same and we're not getting dependency updates there :confused:

PR #171, opened 2025-05-08
allow OKD installs without pullsecret ref

rccrdpccl:
/cc @carbonin 

Create placeholder pullsecret to provide to assisted when not given. OKD installs were requiring pullsecret. Omitting it would result in bootstrap provider failure. OCP installs will fail as the pullsecret will not be valid

Webhook functionality has been added. This will allow to make effectively immutable OpenShiftAssistedConfig: before this change, we could change the resource, but the change would not be applied - and this could cause confusion.
In any case, OpenshiftAssistedConfigTemplate it's still mutable: in case the template is modified, only resources created after the modification will instantiate an OpenshiftAssistedConfig with the new specs.

InfraEnv controller has been removed: it's main purpose was split the logic from the main controller to detect when an ISO was generated. Now we rely on updates on the CreatedTime fields - however it became a really thin controller, merging it with the OpenShiftAssistedConfig controller will make it easier to read and follow the logic.

## Summary by CodeRabbit

- **New Features**
  - Introduced a validating webhook for cluster configuration resources to enforce immutability of specs on updates.
  - Added a cooldown period before fetching ignition data to improve cluster readiness handling.
  - Automatically generates and references a placeholder pull secret when none is provided, ensuring smoother cluster setup.
  - Added webhook TLS certificate management with cert-manager for secure admission control.

- **Bug Fixes**
  - Improved reconciliation logic to prevent premature ignition retrieval and handle pull secret references robustly.

- **Documentation**
  - Updated example YAML templates to remove explicit pull secret declarations and references, simplifying user configuration.

- **Tests**
  - Added webhook validation tests and enhanced existing tests for pull secret handling and cluster reconciliation.

- **Style**
  - Enhanced DNS configuration templates with dynamic logic for different cluster topologies.


bootstrap/internal/controller/openshiftassistedconfig_controller.go
```diff
@@ -223,6 +224,18 @@ func (r *OpenshiftAssistedConfigReconciler) Reconcile(ctx context.Context, req c
 		return ctrl.Result{Requeue: true, RequeueAfter: 60 * time.Second}, nil
 	}
 
+	// ensure there's a pull-secret referenced in the configuration. If none linked by the user, we should generate a fake one
```
carbonin, line 227: hmmm do you think we should do this immediately?
If a user were to just forget the pull secret ref in an OCP install, can they change it later and still have the install succeed?

carbonin: Maybe we could enable this behavior (generating the pull secret) through some kind of flag?

rccrdpccl: That's a very good point.

I've updated the PR to modify the infraenv if the user changes secret ref, however I think it's probably not necessary:
* if bootstrap is ready (node booted with ignition), changing the secret will be useless. We should probably avoid that, maybe add a validation webhook to disallow change if that's the case. We should almost always hit this case
* if bootstrap is not ready, we can update the bootstrap secret (need to do that). This case is hit only if we change the InfraEnv while ISODownloadURL is generating.

As all this logic is in the OpenshiftAssistedConfig controller, race conditions risk on this should be very low (race of changing bootstrap secret values while we're marking at ready)

Question: as ISODownloadURL is not necessary any longer, as soon as infraenv is updated and we hit the ignition endpoint, we should get the updated ignition, right?

In this case we could probably delete the code related to ISODownloadURL waiting, and just implement the validation webhook, thus not allowing infraenv updates, as it should create infraenv and generate ignition in one pass (marking bootstrap as ready).

Regarding the flag, I have mixed feelings: on one hand, it'll definitely help the user save time on errors, on the other hand OKD should be first citizen and it feels awkward having to enable "no pull secret". Maybe we can allow no secret only when detecting is OKD? but that would make it less flexible in case of custom overrides for versions
WDYT?

carbonin: >if bootstrap ...

I'm more thinking about a case where a user is installing OCP so they likely wouldn't be able to pull the images they need once the nodes boot. How do they recover from this after the nodes booted with a fake pull secret and are failing?

>Question: as ISODownloadURL is not necessary any longer, as soon as infraenv is updated and we hit the ignition endpoint, we should get the updated ignition, right?

Technically there will be a gap. Updating the infraenv causes the controller to update the database. Querying for the ignition download queries the database for the ignition data. I suppose there's a chance you update the infraenv and then query for the ignition before the infraenv update goes from etcd -> controller -> db.

But that said, I don't think you have a good way in either case to detect that your specific changes have been applied. The best you can do is watch the infraenv and always reconfigure the host with the new ignition (similar to how BMAC works).

>Maybe we can allow no secret only when detecting is OKD?

From the assisted side I'm not sure we know this at the right time. For example, do we need a pull secret to get any of the images we use during discovery (when we might not know what cluster we're installing to yet)?

rccrdpccl: Summarizing what has been discussed offline:

> I'm more thinking about a case where a user is installing OCP so they likely wouldn't be able to pull the images they need once the nodes boot. How do they recover from this after the nodes booted with a fake pull secret and are failing?

Currently we are in a similar situation. Whatever error in configuration for the bootstrap config (thus infraenv), the user will need to reprovision the nodes.

> But that said, I don't think you have a good way in either case to detect that your specific changes have been applied. The best you can do is watch the infraenv and always reconfigure the host with the new ignition (similar to how BMAC works).

We will watch the infraEnv.Status.CreateTime field, which should be more or less equivalent to ISODownloadURL, but semantically correct.

>From the assisted side I'm not sure we know this at the right time. For example, do we need a pull secret to get any of the images we use during discovery (when we might not know what cluster we're installing to yet)?

We could probably do some basic validation, but for now we'll leave it aside.


Probably the best path would be to treat InfraEnv as immutable, and avoid updating it in the short window we have available: if there' some error, we'll need to delete the resources and apply them again

bootstrap/config/manager/manager.yaml
```diff
@@ -57,5 +57,12 @@ spec:
...
             memory: 64Mi
+        volumeMounts:
+          - mountPath: /tmp/k8s-webhook-server/serving-certs/
+            name: certs
       serviceAccountName: capoa-bootstrap-controller-manager
       terminationGracePeriodSeconds: 10
+      volumes:
+        - name: certs
+          secret:
+            secretName: webhook-cert-secret
```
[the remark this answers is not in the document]
rccrdpccl: https://github.com/openshift-assisted/cluster-api-provider-openshift-assisted/blob/0fd8ac2d2d1537f7aa103e9b59581419bc8eeb3a/bootstrap/config/certmanager/manifests.yaml#L9-L23

bootstrap/internal/controller/openshiftassistedconfig_controller.go
```diff
@@ -273,23 +286,15 @@ func (r *OpenshiftAssistedConfigReconciler) Reconcile(ctx context.Context, req c
...
-			Namespace: machine.Namespace,
-			Name:      infraEnvName,
-		},
-		&infraEnv,
-	); err != nil {
-		return nil, err
-	}
+func isPastIgnitionCooldownPeriod(t time.Time) bool {
+	return time.Now().Add(InfraEnvIgnitionCooldownPeriod).After(t)
+}
```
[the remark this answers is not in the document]
rccrdpccl: good bot

bootstrap/internal/controller/openshiftassistedconfig_controller.go
```diff
@@ -319,18 +324,45 @@ func (r *OpenshiftAssistedConfigReconciler) getIgnitionBytes(ctx context.Context
...
+		return &ie, r.Client.Update(ctx, &ie)
+	}
+
+	// if infraEnv does not exist and does not have pullSecretRef, create it
 	infraEnv := assistedinstaller.GetInfraEnvFromConfig(infraEnvName, config, clusterDeployment)
+	if infraEnv.Spec.PullSecretRef == nil {
+		if err := r.createPullSecretSecretAndRefInfraEnv(ctx, config, infraEnv); err != nil {
+			return infraEnv, err
+		}
+	}
```
carbonin, line 363:
```go
        // check if infraEnv already created
        ie := aiv1beta1.InfraEnv{}

        getInfraEnvErr := r.Get(ctx, types.NamespacedName{Name: infraEnvName, Namespace: config.Namespace}, &ie)

        // return error if the error is not NotFound
        if getInfraEnvErr != nil {
                if !apierrors.IsNotFound(getInfraEnvErr) {
                        return &ie, getInfraEnvErr
                }
                ie = assistedinstaller.GetInfraEnvFromConfig(infraEnvName, config, clusterDeployment)
        }   

        if ie.Spec.PullSecretRef != nil {
                return &ie, nil 
        }

        if err := r.createPullSecretSecretAndRefInfraEnv(ctx, config, &ie); err != nil {
                return nil, err 
        }   

        // if pullsecret ref is not set for an existing infraenv, update it
        if getInfraEnvErr == nil {
                return &ie, r.Client.Update(ctx, &ie)
        }   
```

What do you think about something like this to reduce a bit of the repetition here?

assistedinstaller/auth.go
```diff
@@ -0,0 +1,33 @@
+package assistedinstaller
+
+import (
+	v1 "k8s.io/api/core/v1"
+	v2 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
rccrdpccl: I didn't spot this :facepalm: GoLand has a weird way of aliasing imports

bootstrap/internal/controller/openshiftassistedconfig_controller.go
rccrdpccl: Nice, however we'll need to tweak it as if the infraEnv it's not already created it should be created and not updated, but I think it's simpler than what I have in the PR

carbonin: Right I figured that case would stay the same after what I highlighted. In the case it should be created we'll just continue past what I wrote here.

carbonin: Oh I guess in the case where it returns an infraenv with a pull secret ref we might have a problem :thinking:

rccrdpccl: Thinking about it, I think there's no need to reconcile infraEnv once it has been created. It cannot be created with nil ref to the pull-secret, as it's mandatory, and as the specs of the OpenshiftAssistedConfig are immutable, once we create the infraenv we should be all set. This will simplify the flow and we can get something very similar to your suggestion

bootstrap/internal/controller/openshiftassistedconfig_controller.go
```diff
@@ -461,9 +512,29 @@ func (r *OpenshiftAssistedConfigReconciler) SetupWithManager(mgr ctrl.Manager) e
...
+		).
 		Complete(r)
 }
 
+// Filter infraEnv to be relevant  by this openshiftassistedconfig
+func (r *OpenshiftAssistedConfigReconciler) FilterInfraEnv(ctx context.Context, o client.Object) []ctrl.Request {
+	result := []ctrl.Request{}
+	infraEnv, ok := o.(*aiv1beta1.InfraEnv)
+	if !ok {
+		panic(fmt.Sprintf("Expected an InfraEnv but got a %T", o))
```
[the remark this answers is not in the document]
rccrdpccl: This should never happen, however it's protecting ourselves from programmer's mistake (hooking the function where it shouldn't be hooked)
I don't feel particularly strongly about it: for now I am removing panic but leaving as log - wouldn't mind removing altogether if you feel strongly about it

carbonin (approved):

PR #184, opened 2025-05-20
Revert "Update release candidates"

omer-vishlitzky:
This reverts commit 8c83d99dcdbda6980806c51b2b4c988d0dbd62ac.

As part of the versioning automation efforts, the `version-discovery` job has been modified to create pull requests with updates to the `release-candidates.yaml` file.
Before this modification, it pushed directly to master, but since we had branch protection issues, we tried several other methods to override the branch protection before giving up on it, and one of those attempts pushed an empty `release-candidates.yaml` file in the attached commit.
Hence, reverting this commit.


carbonin:
/lgtm
/approve

Not sure what this is or why it was originally added, but removing an empty file seems harmless to me.
@omer-vishlitzky can you add a bit more detail in the PR description about what is going on here?

PR #194, opened 2025-05-22
ACM-20988: Enable multi-arch konflux builds

carbonin:
Resolves https://issues.redhat.com/browse/ACM-20988


carbonin:
/hold
going to wait for the build to run on the PR then remove the PR check since that's what the team is asking for (multi-arch builds only existing for push jobs)

PR #262, opened 2025-06-16
MGMT-20853: report correctly upgrade status

rccrdpccl:
fix upgrade status reporting in OACP

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved detection of ongoing upgrades to only consider partial updates without a completion time as in progress.

- **Tests**
	- Added new test cases to verify accurate detection of upgrade status in various scenarios.

### DIFF
--- a/controlplane/internal/upgrade/upgrade_test.go
+++ b/controlplane/internal/upgrade/upgrade_test.go
@@ -3,10 +3,14 @@ package upgrade_test
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
 
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
 	"github.com/openshift-assisted/cluster-api-provider-openshift-assisted/controlplane/internal/upgrade"
 	"github.com/openshift-assisted/cluster-api-provider-openshift-assisted/controlplane/internal/workloadclient"
 	"github.com/openshift-assisted/cluster-api-provider-openshift-assisted/pkg/containers"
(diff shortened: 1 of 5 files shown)


controlplane/internal/upgrade/upgrade_test.go
```diff
@@ -186,6 +188,121 @@ var _ = Describe("OpenShift Upgrader", func() {
...
+    history:
+    - acceptedRisks: 'Precondition "ClusterVersionRecommendedUpdate" failed because
+        of "UnknownUpdate": RetrievedUpdates=False (VersionNotFound), so the recommended
+        status of updating from 4.19.0-ec.2 to 4.19.0-ec.5 is unknown.'
+      completionTime: "2025-06-16T11:43:20Z"
+      image: quay.io/openshift-release-dev/ocp-release@sha256:fde51d1c35425759c83ea0895d0d4fc0f88f6a49a0d080207f15a7bbe3e9982c
+      startedTime: "2025-06-16T10:40:09Z"
+      state: Completed
+      verified: true
+      version: 4.19.0-ec.5
```
CrystalChun, line 240: Changes look fine but I wonder if it would be easier and more maintainable to add new tests using the configv1.UpdateHistory struct rather than a yaml string? 
 ```suggestion
    history: []configv1.UpdateHistory{
			{
				State:   configv1.CompletedUpdate,
				Version: "4.19.0-ec.5",
				CompletionTime: "2025-06-16T11:43:20Z",
				Verified: true,
			},
		}

```

rccrdpccl: dates would be something like `meta1v1.Date(2025, 6, 16, 13, 26, 0, UTC)`

PR #334, opened 2025-07-08
Mgmt 20930 expose install config

rccrdpccl:
Expose installConfig override through CAPOA API

## Summary by CodeRabbit

* **New Features**
  * Added support for providing install-config.yaml overrides via annotation on the OpenshiftAssistedControlPlane resource, enabling advanced OpenShift installation customization.
  * Install config overrides are now merged with automatically computed capabilities, ensuring both user and system configurations are respected.

* **Bug Fixes**
  * Ensured that the MachineAPI capability is excluded from baremetal clusters even if specified in overrides.

* **Documentation**
  * Updated getting started guide with instructions on using install config overrides and details about automatic capabilities management.

* **Tests**
  * Enhanced test coverage for install config override annotation handling and capability merging logic.


rccrdpccl:
@carbonin technically we should bump the API, but as it's a backward compatible change (wouldn't break current objects), I would skip it for this - would you agree?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -562,7 +562,11 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensureClusterDeployment(
 			AgentBareMetal: &agent.BareMetalPlatform{},
 		}
 		cd.Spec.PullSecretRef = acp.Spec.Config.PullSecretRef
-
+		if acp.Spec.Config.InstallConfigSecretRef != nil {
+			cd.Spec.Provisioning = &hivev1.Provisioning{
+				InstallConfigSecretRef: acp.Spec.Config.InstallConfigSecretRef,
```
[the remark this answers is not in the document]
rccrdpccl: :facepalm: this is something I've missed

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -325,10 +333,26 @@ func setACICapabilities(oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane
 		aci.Annotations = make(map[string]string)
 	}
 
-	aci.Annotations[InstallConfigOverrides] = string(installCfgOverrideStr)
+	mergedInstallCfgOverride := string(installCfgOverrideStr)
+
+	if hasInstallConfigOverride {
+		mergedInstallCfgOverride = mergeJson(installConfigOverride, mergedInstallCfgOverride)
+	}
+	aci.Annotations[InstallConfigOverrides] = mergedInstallCfgOverride
```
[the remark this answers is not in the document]
rccrdpccl: No, we intend to overwrite the install config with whatever the logic for capabilities decides

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -295,38 +303,87 @@ type InstallConfigOverride struct {
...
 	}
+	return &installConfigOverrideStr, nil
+}
+
+// getInstallConfigOverrideForCapabilities generates install config override JSON for OpenShift capabilities configuration.
+// It automatically sets appropriate capabilities based on the platform type:
+// - Baremetal platforms: Sets baseline to "None" and includes default baremetal capabilities
+// - Non-baremetal platforms: Sets baseline to "vCurrent" and only includes user-specified capabilities
+// Returns nil if no capabilities configuration is needed (non-baremetal with empty capabilities).
+func getInstallConfigOverrideForCapabilities(oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane, aci *hiveext.AgentClusterInstall) (*string, error) {
```
carbonin, line 349: Why make this a string pointer rather than string? I doesn't look like we have different behavior for empty vs nil so feels like a string would just be better.

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -295,38 +303,87 @@ type InstallConfigOverride struct {
...
+// Returns true if the platform type is BareMetalPlatformType, false otherwise.
+func isBaremetalPlatform(aci *hiveext.AgentClusterInstall) bool {
+	return aci.Spec.PlatformType == hiveext.BareMetalPlatformType
+}
+
+// getInstallConfigOverride merges install config override from annotations with capabilities-based overrides.
+// It returns the final merged install config override as a JSON string, or nil if no overrides are needed.
+// The function combines user-provided install config overrides from annotations with automatically
+// generated capabilities configuration based on the cluster platform type.
+func getInstallConfigOverride(oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane, aci *hiveext.AgentClusterInstall) (*string, error) {
```
carbonin, line 316: Same, why use `*string` rather than `string`?

carbonin (commented):
Generally looks better, just some minor comments on types.

controlplane/internal/controller/clusterdeployment_controller.go
```diff
@@ -295,38 +303,87 @@ type InstallConfigOverride struct {
...
+
+// getInstallConfigOverrideForCapabilities generates install config override JSON for OpenShift capabilities configuration.
+// It automatically sets appropriate capabilities based on the platform type:
+// - Baremetal platforms: Sets baseline to "None" and includes default baremetal capabilities
+// - Non-baremetal platforms: Sets baseline to "vCurrent" and only includes user-specified capabilities
+// Returns empty string if no capabilities configuration is needed (non-baremetal with empty capabilities).
+func getInstallConfigOverrideForCapabilities(oacp *controlplanev1alpha2.OpenshiftAssistedControlPlane, aci *hiveext.AgentClusterInstall) (string, error) {
 	var installCfgOverride InstallConfigOverride
-	baselineCapability, err := getBaselineCapability(oacp.Spec.Config.Capabilities.BaselineCapability, isBaremetalPlatform)
+	if isCapabilitiesEmpty(oacp.Spec.Config.Capabilities) && !isBaremetalPlatform(aci) {
```
CrystalChun, line 351: Wouldn't this be either the capabilities is empty or it's not baremetal platform? 

```suggestion
	if isCapabilitiesEmpty(oacp.Spec.Config.Capabilities) || !isBaremetalPlatform(aci) {
```

rccrdpccl: I've preserved current logic: if no capablities and not baremetal, nothing to do and return

otherwise we'll call getBaselineCabality (passing wether it's baremetal or not, and everything should be handled in there).

Am I missing something?

CrystalChun (approved):
/lgtm

PR #610, opened 2026-04-15
Fix DNS resolution: preserve original nameservers in resolv.conf

frajamomo:
The network_setup role was overwriting /etc/resolv.conf with only the libvirt bridge DNS (192.168.222.1) and 8.8.8.8, discarding the original DNS servers from NetworkManager. This broke external DNS resolution (e.g. quay.io) on networks where 8.8.8.8 is blocked.

Now queries NetworkManager for the existing DNS servers and preserves them alongside the libvirt network nameserver.

## Summary by CodeRabbit

* **Tests**
  * DNS server configuration now discovers values from the host system dynamically instead of using hardcoded entries.
  * Generated network definitions and templates now populate upstream DNS entries from the discovered host DNS list rather than fixed addresses.

### DIFF
--- a/test/e2e/libvirt/network-bmh-onlynameserver.xml.j2
+++ b/test/e2e/libvirt/network-bmh-onlynameserver.xml.j2
@@ -13,7 +13,8 @@
     </dhcp>
   </ip>
   <dnsmasq:options>
-    <dnsmasq:option value='server=/registry-proxy.engineering.redhat.com/10.11.5.160'/>
-    <dnsmasq:option value='server=8.8.8.8'/>
+{% for dns in host_dns_servers %}
+    <dnsmasq:option value='server={{ dns }}'/>
+{% endfor %}
   </dnsmasq:options>
 </network>
--- a/test/e2e/libvirt/network-bmh.xml.j2
+++ b/test/e2e/libvirt/network-bmh.xml.j2
@@ -36,7 +36,8 @@
 {% endif %}
     <dnsmasq:option value='address=/assisted-service.assisted-installer.com/{{ loadbalancer_ip }}'/>
     <dnsmasq:option value='address=/assisted-image.assisted-installer.com/{{ loadbalancer_ip }}'/>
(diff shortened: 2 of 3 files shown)


test/playbooks/roles/network_setup/tasks/main.yaml
```diff
@@ -38,10 +38,17 @@
...
+  changed_when: false
+
 - name: Set DNS nameservers in /etc/resolv.conf
   ansible.builtin.copy:
     mode: '0644'
     dest: /etc/resolv.conf
     content: |
+      {% for dns in nm_dns_servers.stdout_lines %}
+      nameserver {{ dns }}
+      {% endfor %}
```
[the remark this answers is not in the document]
rccrdpccl: What the rabbit said is correct. We could add the current DNS config to the dnsmasq config, but right now this solution is incorrect.
However I agree this is an issue, especially when we tear down our test infra and the resolve.conf it's still pointing at an address that is now backed by nothing. 

An alternative could be backing up the current DNS, applying our static dnsmasq, and revert to the original at the end of the t est.

test/e2e/libvirt/network-bmh-onlynameserver.xml.j2
```diff
@@ -14,6 +14,8 @@
   </ip>
   <dnsmasq:options>
     <dnsmasq:option value='server=/registry-proxy.engineering.redhat.com/10.11.5.160'/>
```
rccrdpccl: we can remove this line too

test/e2e/libvirt/network-bmh.xml.j2
```diff
@@ -37,6 +37,8 @@
     <dnsmasq:option value='address=/assisted-service.assisted-installer.com/{{ loadbalancer_ip }}'/>
     <dnsmasq:option value='address=/assisted-image.assisted-installer.com/{{ loadbalancer_ip }}'/>
     <dnsmasq:option value='server=/registry-proxy.engineering.redhat.com/10.11.5.160'/>
```
rccrdpccl: same here, this can be removed

PR #623, opened 2026-04-20
Add cluster-wide TLS adherence support

rccrdpccl:
## Summary

- Adds runtime OpenShift detection via API discovery (`apiservers.config.openshift.io`) and honors cluster-wide TLS adherence policy (`spec.tlsAdherence` / `spec.tlsSecurityProfile` on the `APIServer` resource)
- New `internal/tlsconfig` package resolves TLS configuration for both bootstrap and controlplane managers, with graceful fallback to safe defaults on vanilla Kubernetes
- Wires `SecurityProfileWatcher` (gated behind OCP detection) to trigger manager restart on TLS policy changes
- Bumps `openshift/api`, `assisted-service`, and `controller-runtime` (v0.22.5 → v0.23.3); adapts webhook signatures to new typed `admission.Validator[*T]` generics
- Migrates RBAC to kubebuilder markers for both providers; adds `config.openshift.io/apiservers` permission

## Test plan

- [x] `make test` — all unit tests pass
- [x] `make lint` — 0 issues
- [x] New `internal/tlsconfig` tests cover: vanilla K8s, OpenShift with profile, OpenShift with fetch failure fallback
- [x] e2e on OpenShift cluster with custom TLS profile
- [x] e2e on vanilla Kubernetes cluster (no regression)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Cluster-aware TLS resolution at startup with automatic manager restart on OpenShift TLS policy/profile changes; new TLS resolution utility added.

* **Refactor**
  * Webhook validation converted to typed admission validators and immutability checks simplified.

* **Chores**
  * Expanded controller RBAC for apiservers and agent-related resources; updated dependencies and build contexts to include shared internal sources.

* **Tests**
  * Added TLS resolver unit tests; removed obsolete webhook type-check tests.


internal/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,160 @@
...
+		log.Info("some ciphers from the cluster TLS profile are not supported", "unsupportedCiphers", unsupported)
+	}
+
+	return TLSConfigResult{
+		TLSConfig:          tlsConfig,
+		TLSAdherencePolicy: adherencePolicy,
+		TLSProfileSpec:     profileSpec,
+		IsOpenShift:        true,
+	}, nil
+}
```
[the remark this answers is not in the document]
rccrdpccl: we can't crash the operator on transient API errors

internal/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,158 @@
...
+
+	adherencePolicy, err := crtls.FetchAPIServerTLSAdherencePolicy(ctx, k8sClient)
+	if err != nil {
+		return TLSConfigResult{}, fmt.Errorf("fetching TLS adherence policy from APIServer: %w", err)
+	}
+
+	profileSpec, err := crtls.FetchAPIServerTLSProfile(ctx, k8sClient)
+	if err != nil {
+		return TLSConfigResult{}, fmt.Errorf("fetching TLS profile from APIServer: %w", err)
+	}
```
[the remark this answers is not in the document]
rccrdpccl: You just suggested me the opposite :sweat_smile: 
Anyway, as this is key to be compliant with OpenShift, I think it's better to fail on a transient error rather than badly detecting the platform and fall back on non-compliant TLS policy

internal/tlsconfig/tlsconfig_test.go
```diff
@@ -0,0 +1,142 @@
...
+		})
+
+		It("should return an error when the APIServer resource cannot be fetched", func() {
+			disc := newFakeDiscoveryWithAPIServer()
+			k8sClient := newFakeControllerClient()
+
+			_, err := resolveTLSConfigWithDiscovery(ctx, disc, k8sClient)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("fetching TLS adherence policy from APIServer"))
+		})
```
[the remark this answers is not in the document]
rccrdpccl: According to the compliance rules we should follow APIServer TLS policy. I'd rather fail the container/pod rather than misinterpret it

internal/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,158 @@
...
+		result := buildDefaultResult(true)
+		result.TLSAdherencePolicy = adherencePolicy
+		return result, nil
+	}
+
+	log.Info("honoring cluster-wide TLS profile",
+		"policy", adherencePolicy,
+		"minTLSVersion", profileSpec.MinTLSVersion)
+
+	tlsConfig, unsupported := crtls.NewTLSConfigFromProfile(profileSpec)
```
pastequo, line 104: Should you add a check to ensure the tlsConfig is valid ? For example
- has at least 1 cipher
- the TLS version is consistent with the enable-http2 flag ? (well this one is probably alway true since it's the "min" version and only http 2 can be disabled)

bootstrap/main.go
```diff
@@ -217,8 +231,33 @@ func main() {
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	if tlsResult.IsOpenShift {
+		if err := (&crtls.SecurityProfileWatcher{
+			Client:                    mgr.GetClient(),
+			InitialTLSAdherencePolicy: tlsResult.TLSAdherencePolicy,
+			InitialTLSProfileSpec:     tlsResult.TLSProfileSpec,
+			OnAdherencePolicyChange: func(_ context.Context, oldPolicy, newPolicy configv1.TLSAdherencePolicy) {
+				setupLog.Info("TLS adherence policy changed, shutting down to reload",
```
pastequo: nit: could be a dedicated logger, since it's no longer the setup

bootstrap/main.go
```diff
@@ -217,8 +231,33 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	if tlsResult.IsOpenShift {
+		if err := (&crtls.SecurityProfileWatcher{
```
pastequo: What happens if this one fails at runtime ? Does it fail silently / log the issue and continue / shutdown the whole thing / ... ?

internal/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,158 @@
...
+		TLSAdherencePolicy: adherencePolicy,
+		TLSProfileSpec:     profileSpec,
+		IsOpenShift:        true,
+	}, nil
+}
+
+// hasAPIServerResource checks whether the cluster exposes the
+// apiservers.config.openshift.io resource, which indicates an OpenShift cluster.
+func hasAPIServerResource(disc discovery.DiscoveryInterface) bool {
+	resourceList, err := disc.ServerResourcesForGroupVersion("config.openshift.io/v1")
```
pastequo: nit: this string is probably already the same as the String() of the associated GroupVersion [here](https://github.com/openshift/api/blob/master/config/v1/register.go#L11)

bootstrap/main.go
```diff
@@ -217,8 +231,33 @@ func main() {
...
+			Client:                    mgr.GetClient(),
+			InitialTLSAdherencePolicy: tlsResult.TLSAdherencePolicy,
+			InitialTLSProfileSpec:     tlsResult.TLSProfileSpec,
+			OnAdherencePolicyChange: func(_ context.Context, oldPolicy, newPolicy configv1.TLSAdherencePolicy) {
+				setupLog.Info("TLS adherence policy changed, shutting down to reload",
+					"oldPolicy", oldPolicy, "newPolicy", newPolicy)
+				cancel()
```
pastequo: I guess this stresses the (already existing) fact that the reconciler must be idempotent and that the context can be cancelled any time in the middle of a reconcile loop.

This is not new so this comment might be useless 🙃 
Or maybe it adds some credit about having tests when something is cancelled in the middle of a reconcile loop (if they don't exist already)

pastequo:
Not much to say, I lack context :upside_down_face: , otherwise lgtm

bootstrap/main.go
rccrdpccl: if the profile changes in runtime the app reloads (shutdowns the whole thing): it's intended behaviour to load the new settings.
For transient errors it will retry.

Is this answering your question or did u have any specific scenario in mind?

bootstrap/main.go
rccrdpccl: I am not sure how I would go about testing this: it would need to be tested by a e2e test, and we'd need to perform an action triggering the cancel. How would we possibly spot reliably lack of idempotency?

The aim and intention is for the controller to be idempotent, but bugs might be introduced at some point :sweat_smile: :bug: 

Happy to consider suggestions though!

internal/tlsconfig/tlsconfig.go
rccrdpccl: nice catch, although I am a bit torn on this one I think checking the number of supported ciphers is a good call to comply with the requirements

controlplane/main.go
```diff
@@ -205,8 +232,34 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	if isOpenShift {
+		tlsLog := ctrl.Log.WithName("tls-profile-watcher")
+		if err := (&crtls.SecurityProfileWatcher{
```
jianzzha: bootstrap/main.go and controlplane/main.go contain nearly identical SecurityProfileWatcher setup blocks. Maybe extract it into a shared helper?

bootstrap/main.go
pastequo: it does, indeed it is requeued (I look at the [main branch](https://github.com/openshift/controller-runtime-common/blob/main/pkg/tls/controller.go#L113))

For example, if the object is deleted, there is no callback. Could it be a problem for your case (should you switch back to default) ? Maybe it is not a realistic scenario and this file always exists ?

Feel free to resolve the comment if it's not realistic

rccrdpccl: I believe handling deletion would be more confusing than anything else. APIServer is a resource that it's supposed to be there, and CVO will make sure it is (will reconcile it and recreate it if it was to be deleted).

For our purposes, we wouldn't even know what to do if this object is deleted: what policy should we follow? I think keeping last settings until recreated would be the best choice in this case. Would you agree?

internal/setup/setup.go
```diff
@@ -0,0 +1,64 @@
...
+		InitialTLSProfileSpec:     result.TLSProfileSpec,
+		OnAdherencePolicyChange: func(_ context.Context, oldPolicy, newPolicy configv1.TLSAdherencePolicy) {
+			tlsLog.Info("TLS adherence policy changed, shutting down to reload",
+				"oldPolicy", oldPolicy, "newPolicy", newPolicy)
+			cancel()
+		},
+		OnProfileChange: func(_ context.Context, oldProfile, newProfile configv1.TLSProfileSpec) {
+			tlsLog.Info("TLS profile changed, shutting down to reload",
+				"oldProfile", oldProfile, "newProfile", newProfile)
+			cancel()
```
giladravid16, line 61: Do these logs look good for you? When I tried something like that, there was no space between the args

bootstrap/main.go
pastequo: That might imply to refactor the code to add an orchestration layer. Which would ease the fault injection at any step of the orchestration

Let's keep this conversation for another day

rccrdpccl:
/hold

verifying the logging comment

internal/setup/setup.go
rccrdpccl: I get something like
```
"TLS profile changed, shutting down to reload" logger="tls-profile-watcher" oldProfile={"ciphers":["TLS_AES_128_GCM_SHA256","TLS_AES_256_GCM_SHA384","TLS_CHACHA20_POLY1305_SHA256","ECDHE-ECDSA-AES128-GCM-SHA256","ECDHE-RSA-AES128-GCM-SHA256","ECDHE-ECDSA-AES256-GCM-SHA384","ECDHE-RSA-AES256-GCM-SHA384","ECDHE-ECDSA-CHACHA20-POLY1305","ECDHE-RSA-CHACHA20-POLY1305"],"minTLSVersion":"VersionTLS12"} newProfile={"ciphers":["TLS_AES_128_GCM_SHA256","TLS_AES_256_GCM_SHA384","TLS_CHACHA20_POLY1305_SHA256","ECDHE-ECDSA-AES128-GCM-SHA256","ECDHE-RSA-AES128-GCM-SHA256","ECDHE-ECDSA-AES256-GCM-SHA384","ECDHE-RSA-AES256-GCM-SHA384","ECDHE-ECDSA-CHACHA20-POLY1305","ECDHE-RSA-CHACHA20-POLY1305","ECDHE-ECDSA-AES128-SHA256","ECDHE-RSA-AES128-SHA256","ECDHE-ECDSA-AES128-SHA","ECDHE-RSA-AES128-SHA","ECDHE-ECDSA-AES256-SHA","ECDHE-RSA-AES256-SHA","AES128-GCM-SHA256","AES256-GCM-SHA384","AES128-SHA256","AES128-SHA","AES256-SHA","DES-CBC3-SHA"],"minTLSVersion":"VersionTLS10"}
```

Which is what I've expected and looks decent enough. there are spaces between args, it's just that the profiles are printed without spaces as single objects

PR #629, opened 2026-04-24
Wait for Metal3Cluster CRD before applying cluster manifests

frajamomo:
With CAPI v1.11.3+, the Metal3Cluster CRD may not be fully established by the time the cluster_install role applies the example manifest. This causes a race condition where the Cluster resource fails to create because its infrastructure provider (Metal3Cluster) is not yet recognized by the API server.

Add a retry loop that waits for the Metal3Cluster CRD to reach the Established condition before proceeding with namespace creation and manifest application.

## Summary by CodeRabbit

* **Tests**
  * Enhanced deployment automation to ensure Metal3Cluster prerequisites are fully established before proceeding with component installation, improving deployment reliability and reducing potential race conditions during setup.

### DIFF
--- a/test/playbooks/roles/components_install/tasks/capi_mce.yaml
+++ b/test/playbooks/roles/components_install/tasks/capi_mce.yaml
@@ -388,3 +388,14 @@
   retries: "{{ medium_retries | int }}"
   delay: "{{ medium_delay | int }}"
   when: capm3_standalone
+
+- name: Wait for Metal3Cluster CRD to be established
+  kubernetes.core.k8s_info:
+    api_version: apiextensions.k8s.io/v1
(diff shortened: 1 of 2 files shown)


test/playbooks/roles/cluster_install/tasks/main.yaml
```diff
@@ -1,4 +1,15 @@
 ---
+- name: Wait for Metal3Cluster CRD to be established
```
rccrdpccl: Shouldn't this go together with the metal3 deployment?

frajamomo: Agree, it makes more sense

PR #683, opened 2026-05-13
Add test: transient API error handling

frajamomo:
Add three unit tests for TLS config resolution edge cases:

- NoOpinion ignores configured profile: Verify that resolveOpenShiftTLSConfig uses Intermediate defaults (TLS 1.2) when a Modern profile is explicitly configured on the APIServer CR but tlsAdherence is NoOpinion. Exercises the code path where ShouldHonorClusterTLSProfile returns false despite a restrictive profile being present.
- Transient API error propagation: Verify that a transient API error (ServiceUnavailable) propagates immediately through resolveOpenShiftTLSConfig with no retry or fallback, confirming the current crash-on-transient-error behavior.
- [ACM-34017](https://redhat.atlassian.net/browse/ACM-34017) reproducer (expected failure): Verify that TLSProfileSpec in the result reflects the actual APIServer profile (Modern) when adherence is NoOpinion. This test fails today because resolveOpenShiftTLSConfig stores Intermediate defaults instead of the actual profile. The mismatch causes the SecurityProfileWatcher to detect a spurious profile change on every startup, triggering an infinite restart loop. See [ACM-34017](https://redhat.atlassian.net/browse/ACM-34017).

## Summary by CodeRabbit

* **Tests**
  * Added coverage simulating transient Kubernetes API errors during TLS resolution to verify the transient error is surfaced and only a single API fetch is attempted (no retry/fallback).

[Review Change Stack](https://app.coderabbit.ai/change-stack/openshift-assisted/cluster-api-provider-openshift-assisted/pull/683?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/tlsconfig/tlsconfig_test.go
+++ b/internal/tlsconfig/tlsconfig_test.go
@@ -23,10 +23,12 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	configv1 "github.com/openshift/api/config/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakectrlclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 )
 
 func newFakeControllerClient(objs ...ctrlclient.Object) ctrlclient.Client {
@@ -153,5 +155,30 @@ var _ = Describe("resolveOpenShiftTLSConfig", func() {
 		_, err := resolveOpenShiftTLSConfig(ctx, k8sClient)
 		Expect(err).To(HaveOccurred())
 		Expect(err.Error()).To(ContainSubstring("none of the ciphers from the cluster TLS profile are supported"))
+	})
+})
+
+var _ = Describe("Transient API errors during TLS resolution", func() {
+	It("should propagate a transient API error without retry or fallback", func() {
+		scheme := runtime.NewScheme()
+		_ = configv1.Install(scheme)
+
+		getCalls := 0
+		k8sClient := fakectrlclient.NewClientBuilder().
+			WithScheme(scheme).
+			WithInterceptorFuncs(interceptor.Funcs{
+				Get: func(_ context.Context, _ ctrlclient.WithWatch, _ ctrlclient.ObjectKey, _ ctrlclient.Object, _ ...ctrlclient.GetOption) error {
+					getCalls++
+
+					return apierrors.NewServiceUnavailable("transient API error")
+				},
+			}).
+			Build()
+
+		_, err := resolveOpenShiftTLSConfig(context.Background(), k8sClient)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("fetching TLS adherence policy from APIServer"))
+		Expect(err.Error()).To(ContainSubstring("transient API error"))
+		Expect(getCalls).To(Equal(1), "should call Get exactly once with no retry")
 	})
 })


internal/tlsconfig/tlsconfig_test.go
```diff
@@ -128,3 +130,85 @@ var _ = Describe("resolveOpenShiftTLSConfig", func() {
...
+
+	// ACM-34017: SecurityProfileWatcher restart loop.
+	// When adherence is NoOpinion, resolveOpenShiftTLSConfig returns
+	// Intermediate as TLSProfileSpec (from DefaultTLSConfig) instead of the
+	// actual APIServer profile. The SecurityProfileWatcher is initialized
+	// with this Intermediate spec, but on reconcile it reads the real
+	// APIServer profile (Modern). The mismatch triggers OnProfileChange →
+	// cancel → restart → same mismatch → infinite loop.
+	// ACM-34017: Enable this test once the fix is implemented.
+	PIt("should store the actual APIServer profile in TLSProfileSpec when adherence is NoOpinion", func() {
```
rccrdpccl: Let's either push this with the fix or not push it at all

internal/tlsconfig/tlsconfig_test.go
```diff
@@ -128,3 +130,52 @@ var _ = Describe("resolveOpenShiftTLSConfig", func() {
...
+		k8sClient := newFakeControllerClient(apiServer)
+
+		result, err := resolveOpenShiftTLSConfig(context.Background(), k8sClient)
+		Expect(err).NotTo(HaveOccurred())
+
+		cfg := &tls.Config{}
+		result.TLSConfig(cfg)
+		Expect(cfg.MinVersion).To(Equal(uint16(tls.VersionTLS12)),
+			"NoOpinion should use Intermediate defaults (TLS 1.2), not the configured Modern profile (TLS 1.3)")
+	})
```
rccrdpccl: Is this the expected result? For instance if we set to modern and set back to NoOpinion, should it really go back to Intermediate or stick with Modern?

This test will pass as we're not changing back and forth, but I wonder if we're doing the right assertion. maybe the assertion should be that we won't change what's already set?

frajamomo: The behavior of "fall back to Intermediate when NoOpinion" is CAPOA-specific logic. Other components behave the same.  
IBIO and cluster-capi-operator fall back to Intermediate when adherence is NoOpinion.  The key difference is that IBIO stores the actual APIServer profile in `TLSProfileSpec` regardless of adherence (line: `TLSProfileSpec: tlsProfileSpec`), while CAPOA stores the Intermediate defaults. 
That's the ACM-34017 bug we already identified — CAPOA's approach causes a `SecurityProfileWatcher` mismatch.
The relevant line is in IBIO's `ResolveTLSConfig` return statement:

[tlsconfig.go#L58](https://github.com/openshift/image-based-install-operator/blob/main/internal/tlsconfig/tlsconfig.go#L58)
```
return TLSConfigResult{
    TLSConfig:          tlsConfig,
    TLSAdherencePolicy: tlsAdherencePolicy,
    TLSProfileSpec:     tlsProfileSpec,  // actual APIServer profile, not defaults
}, nil

```
IBIO always stores the real tlsProfileSpec (fetched from APIServer) in TLSProfileSpec, but applies appliedProfile (Intermediate defaults) to TLSConfig. So the watcher is initialized with the real profile and won't see a spurious mismatch on startup — exactly the fix CAPOA needs for ACM-34017.

rccrdpccl: The intention of CAPOA is not to fallback to intermediate when NoOpinion, but rather have a reasonable behaviour where TLSProfileSpec is not present at all (non OCP clusters).

I believe this test is too rigid, and we should test for desired behaviour not current implementation.

frajamomo: In **IBIO** the behavior is:

- TLSConfig = Intermediate (applied to the server)
- TLSProfileSpec = actual APIServer profile (stored for watcher comparison)

Avoids ACM-34017 — the watcher sees the real profile and won't trigger a spurious restart
The server still runs with Intermediate (safe), but the watcher knows what the cluster has.

For non-OCP clusters (kind, vanilla k8s), the APIServer CR doesn't exist at all. CAPOA already handles this — `ResolveOpenShiftTLSConfig` fails to fetch the APIServer, and the caller falls back to `DefaultTLSConfig()` (Intermediate). So the `NoOpinion` path only matters on OCP clusters where the APIServer CR exists but adherence isn't set to `StrictAllComponents`.

The `SecurityProfileWatcher` is initialized with the real profile, not a synthetic one. This prevents the ACM-34017 restart loop — the watcher sees the actual APIServer profile and only triggers restarts on genuine changes.

When an admin switches from `NoOpinion` to `StrictAllComponents`, the watcher's `OnAdherencePolicyChange` callback fires, the manager restarts, and on restart it now honors the profile. The transition is clean because the watcher was already tracking the real profile.

The actual APIServer profile is available but gets discarded when the code takes the NoOpinion path and returns `DefaultTLSConfig()` instead.
```
if !libgocrypto.ShouldHonorClusterTLSProfile(adherencePolicy) {
    result := DefaultTLSConfig()
    result.TLSAdherencePolicy = adherencePolicy
    result.TLSProfileSpec = profileSpec  // store the real profile for the watcher
    return result, nil
}
```

PR #803, opened 2026-08-10
CNV-95021: Reduce memory footprint on multi-namespace clusters

orenc1:
On management clusters with many namespaces (e.g., MCE/ACM environments), the `controller-runtime` informer cache watches all `Roles`, `RoleBindings`, `ClusterRoles`, `ClusterRoleBindings`, `ServiceAccounts`, and `ConfigMaps` cluster-wide. This causes significant memory growth and OOM kills.

Disable caching for these resource types since the controllers access them infrequently and can tolerate direct API reads. Also reduce the default log verbosity from `v=3` to `v=1` in both the controlplane and bootstrap managers to further reduce memory allocation from log formatting.

## Summary by CodeRabbit

* **Performance**
  * Improved control-plane configuration handling by ensuring ConfigMap reads use current Kubernetes data rather than cached values.
* **Documentation**
  * Clarified that the default log level is level 3 (info), making the system’s logging configuration easier to understand.

### DIFF
--- a/bootstrap/main.go
+++ b/bootstrap/main.go
@@ -106,7 +106,7 @@ func main() {
 	initFlags(pflag.CommandLine)
 	pflag.CommandLine.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	// Set log level 2 as default.
+	// Set log level 3 (info) as default.
 	if err := pflag.CommandLine.Set("v", "3"); err != nil {
 		fmt.Println("failed to set default log level", err)
 		os.Exit(1)
--- a/controlplane/main.go
+++ b/controlplane/main.go
@@ -39,6 +39,7 @@ import (
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	"github.com/spf13/pflag"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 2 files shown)


bootstrap/main.go
```diff
@@ -106,8 +106,9 @@ func main() {
 	initFlags(pflag.CommandLine)
 	pflag.CommandLine.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	// Set log level 2 as default.
-	if err := pflag.CommandLine.Set("v", "3"); err != nil {
+	// Set log level 1 (info) as default to reduce memory pressure from verbose
+	// controller-runtime logging on clusters with many namespaces.
+	if err := pflag.CommandLine.Set("v", "1"); err != nil {
```
bluesort: The previous comment was incorrect, but "3" is the value for info-level logging. "1" is warnings and errors only.
See `util/log/log.go`

controlplane/main.go
```diff
@@ -101,8 +104,9 @@ func main() {
 	initFlags(pflag.CommandLine)
 	pflag.CommandLine.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	// Set log level 1 as default.
-	if err := pflag.CommandLine.Set("v", "3"); err != nil {
+	// Set log level 1 (info) as default to reduce memory pressure from verbose
+	// controller-runtime logging on clusters with many namespaces.
+	if err := pflag.CommandLine.Set("v", "1"); err != nil {
```
bluesort: Same as above. This comment was incorrect, but "3" is the right value for info-level logs

controlplane/main.go
```diff
@@ -170,6 +174,18 @@ func main() {
...
 		LeaderElectionID:       "09dff44b.cluster.x-k8s.io",
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
+					&rbacv1.Role{},
+					&rbacv1.RoleBinding{},
+					&rbacv1.ClusterRole{},
+					&rbacv1.ClusterRoleBinding{},
+					&corev1.ServiceAccount{},
+					&corev1.ConfigMap{},
```
bluesort, line 178: Did you check memory or API server logs to confirm the first 5 resources are being watched? Other than `ConfigMap`, I don't see `Get` calls for them in the code

bluesort (changes requested):

controlplane/main.go
orenc1: thanks, that was an oversight from my side, because usually these ones are being cached too. but in this case, the only relevant resource here is the `ConfigMaps`. fixed.

bootstrap/main.go
orenc1: thanks, fixed.
only left the comment fix

controlplane/main.go
orenc1: thanks, fixed.
only left the comment fix

PR #804, opened 2026-08-10
CNV-95022: Prevent upgrade controller from pushing downgrades

orenc1:
## Goal
Enforce that workload cluster upgrades are controller-driven only, with a configurable policy.

### New API field: `spec.upgradePolicy`
- **Managed** (default): the controller actively reverts any externally-initiated upgrades (e.g., admin running `oc adm upgrade` on the workload cluster). Only upgrades driven by updating `spec.distributionVersion` are allowed. This follows the same pattern as HyperShift's HCCO `reconcileClusterVersion`.
- **Unrestricted**: the controller does not block external upgrades. If the workload cluster is upgraded independently, a `VersionDriftDetected` condition is surfaced but no corrective action is taken.

### Upgrade safety logic
- **Never push a downgrade**: if `spec.distributionVersion <= status.distributionVersion`, skip the upgrade push (downgrades are [unsupported by Red Hat](https://access.redhat.com/solutions/4777861))
- **Revert external upgrades** (Managed policy): if an upgrade is in progress that doesn't match `spec.distributionVersion` and the controller is not expecting its own upgrade, clear `ClusterVersion.spec.desiredUpdate`
- **Detect version drift**: if `status > spec` (an external upgrade already completed), surface a `VersionDriftDetected` condition so the cluster admin can update spec to acknowledge
- **Semantic version comparison**: use `semver.ParseTolerant` for version equality checks

### New conditions
| Reason | Meaning |
|--------|---------|
| `ExternalUpgradeBlocked` | An external upgrade attempt was detected and reverted |
| `VersionDriftDetected` | The workload cluster version is ahead of `spec.distributionVersion` |

### New interface method
`ClearDesiredUpdate` nulls the `ClusterVersion.spec.desiredUpdate` field on the workload cluster, following the same pattern as HyperShift's HCCO.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented cluster upgrades from proceeding when the current version is ahead of the requested version.
  * Prevented unintended downgrades and skipped unnecessary upgrade updates.
  * Cleared externally initiated upgrade requests that were not approved by the controller.

* **Improvements**
  * Added clear warning conditions for blocked external upgrades and detected version drift.
  * Upgrade status is now refreshed and retried appropriately when an upgrade is blocked.

### DIFF
--- a/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
+++ b/controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
@@ -293,9 +293,12 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
 
 	var isUpdateInProgress bool
 	var upgradeConditionMessage string
+	var conditionHandled bool
 	defer func() {
+		if conditionHandled {
+			return
+		}
 		if isUpdateInProgress || !isWorkloadClusterRunningDesiredVersion(oacp) {
-			// Either upgrade is in progress or it failed
 			setUpgradeStatus(oacp, isUpdateInProgress, upgradeConditionMessage)
 			return
 		}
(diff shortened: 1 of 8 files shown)


controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -327,6 +327,18 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
...
 
+	// Detect if the tenant cluster upgraded itself past spec.distributionVersion
+	// (e.g., an admin triggered an OTA update from the tenant console). If so,
+	// sync spec up to reflect reality — never push a downgrade.
+	specVersion, specErr := semver.ParseTolerant(oacp.Spec.DistributionVersion)
+	statusVersion, statusErr := semver.ParseTolerant(oacp.Status.DistributionVersion)
+	if specErr == nil && statusErr == nil && statusVersion.GT(specVersion) {
+		log.Info("tenant cluster upgraded independently, syncing spec.distributionVersion",
+			"from", oacp.Spec.DistributionVersion, "to", oacp.Status.DistributionVersion)
+		oacp.Spec.DistributionVersion = oacp.Status.DistributionVersion
```
bluesort, line 338: As a controller, we shouldn't be modifying `spec`. If we detect this case, we should surface a condition so the cluster admin is notified and can update spec themselves

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -346,7 +358,17 @@ func (r *OpenshiftAssistedControlPlaneReconciler) upgradeWorkloadCluster(ctx con
...
-	// once updating, requeue to check update status
+	// Only push an upgrade when spec.distributionVersion is strictly newer than
+	// what the tenant is currently running. This prevents pushing a downgrade
+	// when the versions simply differ (e.g., tenant at 4.16.8, spec at 4.16.2).
+	// If we can't determine the current version, proceed with the push (safe default).
+	if specErr == nil && statusErr == nil && !specVersion.GT(statusVersion) {
+		log.V(logutil.DebugLevel).Info("spec.distributionVersion is not newer than current, skipping upgrade push",
+			"spec", oacp.Spec.DistributionVersion, "status", oacp.Status.DistributionVersion)
+		return ctrl.Result{}, nil
+	}
```
bluesort, line 369: This results in downgrades from spec version being decremented intentionally being ignored

bluesort (changes requested):
Good idea to address version disparity from separate cluster upgrades, but I don't think we can rely on spec vs status versions alone to detect this without blocking downgrades. I also think we should surface a condition once we detect so cluster admins can resolve on their end.

One option is using a new status field tracking the last version successfully applied in `UpdateClusterVersionDesiredUpdate`. We could then detect upgrades external to this controller when status version is different from that field, and spec version can still be used for intended upgrades and downgrades

PR #811, opened 2026-08-14
MGMT-25007: Remove etcd member before control plane machine deletion

bluesort:
Add CAPI pre-terminate hook to remove etcd members from spoke clusters during control plane scale-down, preventing orphaned members that cause permanently degraded etcd.

- Add RemoveEtcdMember to workloadclient.ClientGenerator (port-forward + gRPC to spoke etcd with TLS)
- Set pre-terminate.delete.hook annotation on all control plane Machines
- Add reconcilePreTerminateHook to process deleting machines one at a time, removing etcd member then clearing the hook
- Filter deleting machines in reconcileReplicas via ActiveMachines
- Fix wrong "creating" log message on scale-down path

## Summary by CodeRabbit

* **New Features**
  * Added automatic etcd member cleanup when control-plane Machines are deleted.
  * Added orphaned etcd member detection and removal.
  * Added etcd leadership transfer before member removal or scale-down.
  * Added safeguards to process deletions sequentially and protect cluster health.
  * Added cleanup and restoration of termination hooks for control-plane Machines.
  * Added termination-hook cleanup during control-plane deletion.

* **Bug Fixes**
  * Improved replica calculations by excluding Machines already being deleted.
  * Added handling for missing node references, provisioning states, and transient cleanup failures.

* **Tests**
  * Added comprehensive coverage for etcd operations, hooks, retries, errors, and deletion scenarios.


controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -527,10 +572,69 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensureClusterDeployment(
...
+		return ctrl.Result{}, nil
+	}
+
+	machinesWithHook := deletingMachines.Filter(collections.HasAnnotationKey(preTerminateHookAnnotation))
+	machinesWithoutHook := deletingMachines.Difference(machinesWithHook)
+
+	if machinesWithoutHook.Len() > 0 {
+		log.V(logutil.DebugLevel).Info("waiting for machines to complete deletion before processing next pre-terminate hook", "count", machinesWithoutHook.Len())
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
```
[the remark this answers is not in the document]
bluesort: This is best as a followup issue if it ends up being a concern

rccrdpccl:
@bluesort please next time separate commits between the actual change and the vendor, it will make it easier to review - especially on GH UI

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -527,10 +572,69 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensureClusterDeployment(
...
+
+	deletingMachines := allMachines.Filter(collections.HasDeletionTimestamp)
+	if deletingMachines.Len() == 0 {
+		return ctrl.Result{}, nil
+	}
+
+	machinesWithHook := deletingMachines.Filter(collections.HasAnnotationKey(preTerminateHookAnnotation))
+	machinesWithoutHook := deletingMachines.Difference(machinesWithHook)
+
+	if machinesWithoutHook.Len() > 0 {
```
rccrdpccl: if a machine is stuck in terminating, we'll be requeuing every 30s and never reconcile other machines, am I understanding correctly? If so, one stuck machine might freeze other controlplane reconciliations.

controlplane/internal/workloadclient/etcd.go
```diff
@@ -0,0 +1,249 @@
...
+		if member.Name == memberName {
+			targetMemberID = member.ID
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		return nil
+	}
```
rccrdpccl: not found has the same output of removed? Is this intentional?

rccrdpccl (changes requested):
Can we add functionality to handle already provisioned clusters?

controlplane/internal/workloadclient/etcd.go
bluesort: yes, intentional for idempotency - do we typically handle this in a different way?

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
bluesort: true. we should simply proceed after retrying for a certain timeframe

bluesort: hm, actually I don't think need to block on deleting machines at all. once the hook is removed the host has already been deleted in etcd and we can let CAPI worry about tearing down

controlplane/internal/workloadclient/etcd.go
rccrdpccl: I think in general it's a good pattern, but could we run into race conditions on edge cases? For instance how would spinning up and quickly down a node work? Could we risk ending up with the same situation? For instance, would it be possible that a node booted but didn't register yet, and registers before we actually clean it up?

rccrdpccl: I think we should probably add orphan pruning, as it's done in [KCP](https://github.com/kubernetes-sigs/cluster-api/blob/94a0ca7ef3760745007b06bf63820a8358b8dec2/controlplane/kubeadm/reconcilers/kubeadmcontrolplane/kubeadmcontrolplane_controller.go#L488-L491)
We could then even release on error (when removeEtcdMember fails) instead of requeuing, relying that eventually we'll be pruning etcd members. This will allow to carry on reconciling

I think we still have some other gaps, compared to KCP implementation:
* removing nodes that [would break quorum](https://github.com/kubernetes-sigs/cluster-api/blob/94a0ca7ef3760745007b06bf63820a8358b8dec2/controlplane/kubeadm/reconcilers/kubeadmcontrolplane/kubeadmcontrolplane_controller.go#L1493-L1496) 
* [forwarding leadership](https://github.com/kubernetes-sigs/cluster-api/blob/94a0ca7ef3760745007b06bf63820a8358b8dec2/controlplane/kubeadm/reconcilers/kubeadmcontrolplane/kubeadmcontrolplane_controller.go#L1503-L1508)

bluesort: Orphan pruning is a good idea. Looks like KCP uses a pre-commit hook as best-effort, but removes the hook if a machine can't be processed correctly, at which point the machine is removed from the hub cluster and orphan pruning will clean it up in the spoke cluster etcd at next run. I've updated to match that behavior here and cover the edge cases we discussed

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -527,10 +582,180 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensureClusterDeployment(
...
+		orphanName = "(name not yet assigned)"
+	}
+	log.V(logutil.InfoLevel).Info("removing orphaned etcd member", "memberName", orphanName, "memberID", orphan.ID)
+
+	if err := r.WorkloadClientGenerator.RemoveEtcdMemberByID(ctx, kubeconfig, orphan.ID); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to remove orphaned etcd member %s (id: %d): %w", orphanName, orphan.ID, err)
+	}
+
+	return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
+}
```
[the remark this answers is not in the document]
bluesort: No etcd members are removed while a machine is actively provisioning. Considering a member an orphan when there is no active machine tied to it is fine semantically, it will eventually be claimed by the provisioning machine (or not, in which case it will be pruned)

controlplane/internal/controller/openshiftassistedcontrolplane_controller.go
```diff
@@ -527,10 +582,180 @@ func (r *OpenshiftAssistedControlPlaneReconciler) ensureClusterDeployment(
...
+		}
+	}
+	return names
+}
+
 func (r *OpenshiftAssistedControlPlaneReconciler) reconcileReplicas(ctx context.Context, oacp *controlplanev1alpha3.OpenshiftAssistedControlPlane, cluster *clusterv1.Cluster) error {
 	log := ctrl.LoggerFrom(ctx)
 	ownerGK := schema.GroupKind{Group: controlplanev1alpha3.Group, Kind: "OpenshiftAssistedControlPlane"}
-	machines, err := collections.GetFilteredMachinesForCluster(ctx, r.Client, cluster, collections.OwnedMachines(oacp, ownerGK))
+	machines, err := collections.GetFilteredMachinesForCluster(ctx, r.Client, cluster, collections.OwnedMachines(oacp, ownerGK), collections.ActiveMachines)
```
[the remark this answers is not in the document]
bluesort: This is a pre-existing issue not in scope here

rccrdpccl:
great addition!!

/lgtm
/approve

PR #827, opened 2026-08-23
CNV-95786: Add KubeVirt platform detection and infra cluster client

orenc1:
Add utility functions for KubeVirt platform detection from the CAPI Cluster's `infrastructureRef`, and a client builder for connecting to external infrastructure clusters.

Platform detection:
- `IsKubeVirtInfra()` checks `Cluster.spec.infrastructureRef.Kind`

Infra cluster client:
- Reads kubeconfig from Secret referenced by `spec.config.infraClusterRef`
- Falls back to local client when no ref is configured (same-cluster topology)
- Determines infra namespace from kubeconfig context

API addition:
- Add `infraClusterRef` field to `OpenshiftAssistedControlPlaneConfigSpec`


controlplane/internal/kubevirt/infra_client.go
```diff
@@ -0,0 +1,125 @@
...
+) (*InfraClientResult, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	if !IsKubeVirtInfra(cluster) {
+		return nil, fmt.Errorf("cluster %s is not a KubeVirt cluster", cluster.Name)
+	}
+
+	// Fetch the KubevirtCluster resource
+	kvCluster := &unstructured.Unstructured{}
+	kvCluster.SetGroupVersionKind(schema.GroupVersionKind{
```
rccrdpccl: We could add `infraClusterRef` or something to the user-faced API in CAPOA, this way we would completely decouple from kuebevirt provider and its running implementation. The downside is that the user will need to type this possibly twice (one for kubevirt infra provider, one for CAPOA)

orenc1: This makes a total sense if we want a true decoupling from CAPK.
originally this field was in the PR https://github.com/openshift-assisted/cluster-api-provider-openshift-assisted/pull/819, but i removed it so the API additions will be as minimal as possible.
The PR has been updated and now `GetInfraClusterClient()` reads the kubeconfig from OACP's `.spec.config.infraClusterRef`

controlplane/api/v1alpha3/openshiftassistedcontrolplane_types.go
```diff
@@ -57,6 +57,14 @@ type OpenshiftAssistedControlPlaneSpec struct {
 type OpenshiftAssistedControlPlaneConfigSpec struct {
 	// From AgentClusterInstall https://github.com/openshift/assisted-service/blob/master/api/hiveextension/v1beta1/agentclusterinstall_types.go
 
+	// InfraClusterRef references a Secret containing the kubeconfig for the
+	// infrastructure cluster where workload VMs and supporting resources
+	// (Services, Routes, DNS proxy, golden PVC) are created.
+	// If not set, the controller assumes the management cluster IS the infra
+	// cluster (local-infra topology).
+	// +optional
+	InfraClusterRef *corev1.LocalObjectReference `json:"infraClusterRef,omitempty"`
```
rccrdpccl: Would we ever need this for bootstrapping worker nodes?

orenc1: no, the bootstrap controller doesn't need the infra cluster credentials, because it just generates ignition/bootstrap configs (discovery ignition for CP nodes, MCS pointer ignition for workers), then writes the config to a secret that CAPK reads, and that's it. no infra cluster interaction.
only CAPOA control plane controller and CAPK are interacting with the infra cluster.

rccrdpccl (approved):

